PR #59, opened 2025-03-20
[ExampleApp] Implement Model Selection Screen and downloader

leehack:
The goal of this is to make it similar to https://github.com/google-ai-edge/mediapipe-samples/tree/main/examples/llm_inference/android

- [x] Implement Model selection UI
- [x] Disable CPU models on the Web
- [x] Add HuggingFace Authentication
- [x] Add HuggingFace Model License Agreement
- [x] Implement Model Downloader
- [x] Make it use the downloaded model
- [x] Add Web support or Disable it at least

The model downloader is only supported for mobile devices.
For the web, the model can be served from the local assets as usual.

image
image


leehack:
Test result.
Web: Only work with local model. (Tested with Gemma3)
Android: Gemma3 GPU/CPU and DeepSeek
iOS: DeepSeek. (Gemma3 doesn't work)

example/lib/chat_screen.dart
```diff
@@ -27,26 +29,29 @@ class ChatScreenState extends State<ChatScreen> {
     _initializeModel();
   }
 
+  @override
+  void dispose() {
+    super.dispose();
+    _gemma.modelManager.deleteModel();
```
DenisovAV, line 35: I don't think we need delete model, because in this case you need to download it again during the next launch. As an option we can ask user, if they would like to remove model, and remove in case of agreement

example/lib/model_selection_screen.dart
```diff
@@ -0,0 +1,57 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_gemma/pigeon.g.dart';
```
DenisovAV, line 3: I would propose to use names like Gemma 3 GPU, Gemma 3 CPU, Deepseek, Gemma, not just default strings from enum

example/lib/chat_screen.dart
leehack: When a user gets back to this screen without this, it crashes. It's because the _gemma is a singleton. Since the example now only uses [setModelPath](https://github.com/DenisovAV/flutter_gemma/pull/59/files#diff-4fda3d60335fd0ae6fc1d5c79c4a0e85c26a205d04f5b5ecb9b67520ef5bbaR43), it doesn't redownload.

Maybe we could improve the initialize logic so that it doesn't need to re-init anything if it has a model created and the model hasn't changed.

example/lib/model_selection_screen.dart
```diff
@@ -0,0 +1,57 @@
...
+import 'package:flutter_gemma_example/chat_screen.dart';
+import 'package:flutter_gemma_example/model_download_screen.dart';
```
DenisovAV, line 5: And I think we need to provide the first question if model is not dowloaded yet "Download Model / Asset Model" and if user choose Download - navigate they to Download screen

example/lib/chat_screen.dart
DenisovAV: Got it! Let's keep this fix for now, I'll fix init after merge and remove

example/lib/model_selection_screen.dart
leehack: I'll work on that tomorrow :)

example/lib/model_selection_screen.dart
leehack: I think the asset model isn't recommended unless the developer really wants to do it. Even the smallest model is over 500mb(Gemma 3 1b). The current flow isn't clear for web tho. Technically there are no assets anymore since the web's flow is actually just downloading it from the static server.

example/lib/model_selection_screen.dart
leehack: done

example/lib/model_selection_screen.dart
DenisovAV: Yes, but I still suggest keeping this option, as it's the only way to download the model once and use it across all platforms. Here's the scenario I propose:

The user selects either "Download" or "Assets".

- If they choose "Download", it works the same way as it does now.

- If they choose "Assets", we ask which model is included in the assets and allow them to select it.

For the web, the "Download" option won’t be available. On mobile, we should display a clear message in bold that the "Asset" mode is for debug purposes only.

leehack: Just an idea. So, currently, the model select list shows the local model only for web.
What if we just show it for the mobile as well? (Probably I could also check if the local model files are even there if it's mobile)
What do you think?

DenisovAV: I’ve got an another idea! Let’s not use assets at all. Instead, in addition to the predefined models in the list, let’s add an option to manually enter the path to the model and select its type. That way, users can point to any model they want. For web, we should keep only manual path. What do you think?

leehack: That works. So, instead of having a local model in the model list, we add an option for `Custom model`? When a user selects the option, there's a screen that allows the user to configure the path and model configs, right?

I like the idea. Actually, it should probably be `Add a model` instead of `Custom model`. This way, user can persist their config and can be shown in the list once added.

leehack: Btw, I feel this can be implemented in another PR. Haha.

PR #179, opened 2026-01-31
fix: use local app data directory on desktop platforms

ProjectEdge-Jim:
## Summary

On desktop platforms, the Documents folder is often synced to cloud services (OneDrive, iCloud, Dropbox). Large model files (~3.6GB) in these folders cause issues with native code file access - specifically, LiteRT-LM's JNI layer cannot reliably access files in cloud-synced locations.

This change uses platform-specific local app data directories that are never synced:

| Platform | Path | Notes |
|----------|------|-------|
| Windows | `%LOCALAPPDATA%\flutter_gemma` | Truly local, never synced (unlike Roaming) |
| macOS | `~/Library/Application Support/flutter_gemma` | Not synced by default |
| Linux | `~/.local/share/flutter_gemma` | Respects `XDG_DATA_HOME` |
| Android/iOS | Documents directory | Unchanged |

## Problem

When a Windows user has OneDrive enabled (which syncs the Documents folder), downloading a Gemma model via `getApplicationDocumentsDirectory()` places the ~3.6GB file in a cloud-synced location. The LiteRT-LM native JNI code then fails to load the model with:

```
com.google.ai.edge.litertlm.LiteRtLmJniException: Model file not found: C:\Users\...\OneDrive\Documents\gemma-3n-E2B-it-int4.litertlm
```

The file exists but native code cannot access it through the OneDrive virtualized path.

## Solution

Use `%LOCALAPPDATA%` on Windows instead of `getApplicationDocumentsDirectory()`. This directory:
- Is never synced by OneDrive or any other cloud service
- Is the standard location for app-specific local data on Windows
- Works reliably with native JNI code

Similar platform-appropriate directories are used on macOS and Linux.

## Testing

- Tested on Windows 11 with OneDrive enabled
- Model downloads to `C:\Users\<user>\AppData\Local\flutter_gemma\`
- LiteRT-LM loads the model successfully

## Breaking Changes

None. Mobile platforms (Android/iOS) continue to use the Documents directory as before. Existing desktop users will need to re-download the model to the new location.

### DIFF
--- a/lib/core/infrastructure/platform_file_system_service.dart
+++ b/lib/core/infrastructure/platform_file_system_service.dart
@@ -132,7 +132,18 @@ class PlatformFileSystemService implements FileSystemService {
     // The actual tracking is done in ProtectedFilesRegistry.registerExternalPath()
   }
 
-  /// Gets the app documents directory with caching
+  /// Gets the model storage directory with caching.
+  ///
+  /// On desktop platforms, uses platform-specific local app data directories
+  /// that are NOT synced to cloud services (OneDrive, iCloud, Dropbox).
+  /// This is critical because native code (LiteRT-LM JNI) cannot reliably
+  /// access files in cloud-synced folders.
+  ///
+  /// Storage locations:
+  /// - Windows: %LOCALAPPDATA%\flutter_gemma (truly local, never synced)
+  /// - macOS: ~/Library/Application Support/flutter_gemma
+  /// - Linux: ~/.local/share/flutter_gemma
+  /// - Android/iOS: App documents directory (standard behavior)
   Future<Directory> _getDocumentsDirectory() async {
     // Web doesn't support local file system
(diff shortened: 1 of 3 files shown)


DenisovAV:
Closing — same fix landed in 0.15.1 (commits 9c42c4a + f828752), with your co-author credit. We extended your approach with a read-from-both fallback so existing installs on `Documents/` are auto-detected instead of forcing a re-download — your "Breaking change" note no longer applies. Paths chosen match yours exactly:

- Windows: `%LOCALAPPDATA%\flutter_gemma`
- macOS: `~/Library/Application Support/flutter_gemma`
- Linux: `~/.local/share/flutter_gemma`
- Mobile: unchanged

Thanks for the diagnosis and PR — the OneDrive-virtualized-path breakage on JNI/FFI was real and your fix shape was exactly right. Will ship in 0.15.1.

PR #230, opened 2026-04-20
feat: make maxFunctionBufferLength configurable in createChat()

### LINKED ISSUE #229: Make _maxFunctionBufferLength configurable in createChat()
## Problem

When using function calling with models that produce verbose tool call output (e.g. Gemma 4 E4B sharing recipe content), the function call buffer limit of 1024 characters (`_maxFunctionBufferLength` in `lib/core/chat.dart:16`) causes the response to be flushed as plain text instead of being parsed as a `FunctionCallResponse`.

This means any tool call whose raw text exceeds 1024 characters is silently downgraded to a `TextResponse`, with no way for the caller to adjust the threshold.

## Where the limit is enforced

```dart
// lib/core/chat.dart:16
const int _maxFunctionBufferLength = 1024;
```

This constant controls how many characters the streaming parser accumulates before giving up and flushing the buffer as text. It is not exposed as a parameter.

## Why it matters

Real-world function calls can legitimately exceed 1024 characters. For example, a `share_recipe` tool call where the model includes a full recipe in the `content` argument easily reaches 1500-2000+ characters. The current behavior silently breaks function calling for these use cases with no error or warning.

Workarounds like instructing the model to shorten its output are fragile and limit what tools can do.

## Proposed solution

Expose `maxFunctionBufferLength` as an optional parameter in `createChat()` with the current `1024` as default (no breaking change):

```dart
Future<InferenceChat> createChat({
  // ... existing parameters ...
  int maxFunctionBufferLength = 1024, // NEW
}) async {
```

And pass it through to `InferenceChat`:

```dart
InferenceChat({
  // ... existing parameters ...
  this.maxFunctionBufferLength = 1024,
});
```

Then use it instead of the hardcoded constant in the streaming logic.

## Environment

- flutter_gemma: 0.13.5
- Model: Gemma 4 E4B (gemma-4-E4B-it.litertlm)
- Platform: Android (LiteRT-LM backend)

using-system:
## Summary

- Expose `maxFunctionBufferLength` as an optional parameter in `createChat()` (default `1024`, no breaking change)
- Replace the hardcoded `_maxFunctionBufferLength` constant with a per-chat instance field
- Applied to both the base `InferenceModel` and the `DesktopInferenceModel` override

## Problem

The hardcoded `_maxFunctionBufferLength = 1024` in `lib/core/chat.dart` causes function calls with content exceeding 1024 characters to be silently flushed as plain `TextResponse` instead of being parsed as `FunctionCallResponse`.

This breaks real-world use cases like recipe sharing where the model includes formatted content in tool call arguments (easily 1500-2000+ characters).

## Usage

```dart
final chat = await model.createChat(
  tools: myTools,
  supportsFunctionCalls: true,
  maxFunctionBufferLength: 4096, // increase for verbose tool calls
);
```

Closes #229

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/core/chat.dart
+++ b/lib/core/chat.dart
@@ -10,10 +10,9 @@ import 'package:flutter_gemma/flutter_gemma_interface.dart';
 
 import 'model.dart';
 
-// Constants
-/// Maximum length for function call buffer before flushing as text.
+/// Default maximum length for function call buffer before flushing as text.
 /// Must accommodate verbose formats (DeepSeek tags, parallel calls).
-const int _maxFunctionBufferLength = 1024;
+const int defaultMaxFunctionBufferLength = 1024;
 
 class InferenceChat {
   final Future<InferenceModelSession> Function()? sessionCreator;
@@ -22,6 +21,7 @@ class InferenceChat {
   final bool supportImage;
   final bool supportAudio;
   final bool supportsFunctionCalls;
+  final int maxFunctionBufferLength;
   final ModelType modelType; // Add modelType parameter
   final bool isThinking; // Add isThinking flag for thinking models
   final ModelFileType fileType; // Add fileType parameter
@@ -46,6 +46,7 @@ class InferenceChat {
     this.supportImage = false,
     this.supportAudio = false,
     this.supportsFunctionCalls = false,
+    this.maxFunctionBufferLength = defaultMaxFunctionBufferLength,
     this.tools = const [],
     this.modelType =
         ModelType.gemmaIt, // Default to gemmaIt for backward compatibility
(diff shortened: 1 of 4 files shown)


DenisovAV (changes requested):
Thanks for the PR! The approach is correct and the problem is real.

One issue: `MobileInferenceModel.createChat()` in `lib/mobile/flutter_gemma_mobile_inference_model.dart` is missing the `maxFunctionBufferLength` parameter. Currently the feature only works on Desktop — on Android/iOS the parameter is silently ignored.

Please add:
1. `int? maxFunctionBufferLength,` to the mobile `createChat()` signature (line 34)
2. `maxFunctionBufferLength: maxFunctionBufferLength ?? defaultMaxFunctionBufferLength,` in the `InferenceChat` constructor call (around line 52)

After that it's good to merge.

using-system:
@DenisovAV Fixed in f633e90. Added `int? maxFunctionBufferLength` to `MobileInferenceModel.createChat()` signature and forwarded it to the `InferenceChat` constructor with `defaultMaxFunctionBufferLength` as fallback — mirroring the Desktop implementation. The parameter now works consistently across all platforms.

Thanks for the review!

DenisovAV (approved):

PR #237, opened 2026-04-25
fix(desktop): inherit parent process environment in server subprocess

RZI3D:
## Checklist
- [x] Tested on Linux (NixOS) with flutter_gemma 0.13.6
- [x] Does not break existing behavior on standard distros
- [x] Single file change, no new dependencies
- [x] Follows existing code style

## Problem

On non-standard Linux distributions (NixOS mostly.), the LiteRT-LM 
server subprocess fails to initialize the native engine with a silent 
gRPC UNKNOWN error (code 2).

The root cause is that `Process.start()` is called with a manually 
constructed environment that only contains `LD_LIBRARY_PATH` set to 
the natives path — completely replacing the parent process environment. 
This means any system libraries not in standard paths (such as 
`libvulkan.so.1` on NixOS) are invisible to the subprocess, causing 
`liblitertlm_jni.so` to fail loading at runtime with no useful error message.

## Fix

Spread `Platform.environment` into the subprocess environment first, 
then append the natives path to `LD_LIBRARY_PATH`/`DYLD_LIBRARY_PATH` 
rather than replacing it. This ensures the subprocess inherits all 
parent environment variables while still prioritizing the bundled natives.

## Testing

Verified on NixOS with `flutter_gemma` 0.13.6 — model initializes 
successfully after this change. Previously failed silently at 
`Creating EngineConfig` with `gRPC Error (code: 2, codeName: UNKNOWN, 
message: null)`.

## Notes

This is a one-line fix that shouldn't affect any existing behavior on 
standard distros — if `LD_LIBRARY_PATH` is already set correctly, 
spreading `Platform.environment` and appending to it is equivalent 
to the previous behavior.

### DIFF
--- a/lib/desktop/server_process_manager.dart
+++ b/lib/desktop/server_process_manager.dart
@@ -116,8 +116,8 @@ class ServerProcessManager {
           _currentPort.toString(),
         ],
         environment: {
-          if (Platform.isLinux) 'LD_LIBRARY_PATH': nativesPath,
-          if (Platform.isMacOS) 'DYLD_LIBRARY_PATH': nativesPath,
+          if (Platform.isLinux) 'LD_LIBRARY_PATH': '$nativesPath:${Platform.environment['LD_LIBRARY_PATH'] ?? ''}',
+          if (Platform.isMacOS) 'DYLD_LIBRARY_PATH': '$nativesPath:${Platform.environment['DYLD_LIBRARY_PATH'] ?? ''}',
         },
       );
 


DenisovAV:
Thanks for the detailed report and the fix!

Heads-up though: in a couple of days I'm landing a major rewrite that drops the JVM gRPC server entirely and switches desktop to a direct dart:ffi call into the LiteRT-LM C API (see the `feature/desktop-ffi` branch). It will be noticeably faster and `Process.start()` goes away completely — so this fix won't survive the rewrite.

Given that, I'd rather not merge this PR. The NixOS issue will resolve itself once the FFI version ships, since `dlopen` in the current process will pick up your environment naturally.

Really appreciate you tracking this down — closing as superseded by the upcoming FFI work.

RZI3D:
Ok, sounds good! The java server was being a pain anyways, it for some reason would never quit and keep running in the background. Ima keep an eye on `feature/desktop-ffi`. Thank you for your time

PR #288, opened 2026-05-21
feat: report initialized backend

merlinnot:
## Summary
- expose `InferenceModel.activeBackend` so callers can inspect the backend actually initialized by the runtime
- make FFI `.litertlm` initialization try backend fallback inside the plugin and return the backend that succeeded
- share and test the FFI backend fallback order / LiteRT-LM wire names
- report null for web and MediaPipe runtimes where the final initialized backend is not currently observable

## Motivation
Callers can request a preferred backend, but the runtime may fall back. This gives applications a way to show users the compute backend actually in use without implementing their own fallback or probe loop.

## Tests
- `test/core/ffi/backend_preference_test.dart` covers NPU -> GPU -> CPU, GPU -> CPU, null -> GPU -> CPU, CPU-only fallback, and wire names
- `test/core/ffi/ffi_inference_model_test.dart` covers that the FFI model exposes the initialized backend

## Validation
- flutter pub get
- dart analyze lib/flutter_gemma_interface.dart lib/core/ffi/backend_preference.dart lib/core/ffi/ffi_inference_model.dart lib/core/ffi/ffi_inference_model_stub.dart lib/mobile/flutter_gemma_mobile_inference_model.dart lib/web/flutter_gemma_web.dart lib/mobile/flutter_gemma_mobile.dart lib/desktop/flutter_gemma_desktop.dart test/core/ffi/backend_preference_test.dart test/core/ffi/ffi_inference_model_test.dart
- flutter test test/core/ffi/backend_preference_test.dart test/core/ffi/ffi_inference_model_test.dart test/core/litertlm_file_type_test.dart test/desktop_vision_params_test.dart --reporter=failures-only
- git diff --check

## Notes
`flutter pub get` updated local lockfile resolution for `meta` and `test_api` with this Flutter SDK, but those lockfile changes were intentionally not included in the commit.

### DIFF
--- a/lib/core/ffi/backend_preference.dart
+++ b/lib/core/ffi/backend_preference.dart
@@ -1,0 +1,23 @@
+import '../../pigeon.g.dart';
+
+List<PreferredBackend> ffiBackendFallbackOrder(
+  PreferredBackend? preferredBackend,
+) =>
+    switch (preferredBackend) {
+      PreferredBackend.npu => const [
+          PreferredBackend.npu,
+          PreferredBackend.gpu,
+          PreferredBackend.cpu,
+        ],
+      PreferredBackend.gpu || null => const [
+          PreferredBackend.gpu,
+          PreferredBackend.cpu,
+        ],
+      PreferredBackend.cpu => const [PreferredBackend.cpu],
+    };
+
+String ffiBackendWireName(PreferredBackend backend) => switch (backend) {
+      PreferredBackend.npu => 'npu',
+      PreferredBackend.gpu => 'gpu',
+      PreferredBackend.cpu => 'cpu',
+    };
--- a/lib/core/ffi/backend_preference.dart
+++ b/lib/core/ffi/backend_preference.dart
@@ -24,6 +24,7 @@ String ffiBackendWireName(PreferredBackend backend) => switch (backend) {
       PreferredBackend.cpu => 'cpu',
     };
 
+/// Failure details for a single FFI backend initialization attempt.
 class BackendInitAttemptFailure {
   const BackendInitAttemptFailure({
     required this.backend,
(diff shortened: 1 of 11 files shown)


DenisovAV:
Thanks for this — the activeBackend getter + automatic fallback chain
is exactly the observability hook that's been missing. The pure-function
extraction in backend_preference.dart reads well, tests are well-scoped,
and the diff rebases cleanly on top of release/0.16.2.

Three things I'd like to address before merging, all in the fallback
loop in flutter_gemma_desktop.dart:387 / flutter_gemma_mobile.dart:692:

1. Throw the last error, not the first — or aggregate. Today
   `firstError ??= error` means a user who requested NPU and effectively
   ran on CPU sees the NPU error, not whatever made CPU fail (if it
   did). When all backends fail, the last attempt is the configuration
   they actually got — it's usually the most actionable. Even better:
   a small BackendInitException(attempts: [...]) so the chain is
   inspectable.

2. `on Object catch` is wider than intended. It traps OutOfMemoryError,
   StackOverflowError, AssertionError etc., which are programming bugs
   rather than backend-capability problems. Either `on Exception catch`,
   or `on Object catch (e) { if (e is OutOfMemoryError || e is
   StackOverflowError) rethrow; ... }`.

3. Silent NPU → CPU fallback in release builds. debugPrint is
   debug-only — in a packaged app the user gets a CPU model and has no
   visible signal that NPU failed unless they explicitly read
   activeBackend. Could you switch the fallback log to
   developer.log(level: 900 /* WARNING */, name: 'flutter_gemma') so
   it survives release builds? And maybe note the silent-fallback
   semantics in the activeBackend dartdoc so callers know they're
   expected to check.

Smaller items, fine to fold into the same revision or leave for later:

- The two _initializeFfiInferenceRuntime helpers in desktop and mobile
  are byte-identical except for the log tag — could live in
  core/ffi/backend_preference.dart with the tag as a parameter.
- The firstError! bang at the end of each helper is logically safe
  (the order list is always non-empty) but a small `if (firstError ==
  null) throw StateError(...)` would make that invariant explicit and
  survive future enum edits.

Sasha

merlinnot:
Thanks for the review, Sasha. I addressed the fallback loop feedback in the latest commits:

- FFI backend initialization now uses a shared fallback helper for desktop/mobile.
- Failed fallback chains now throw `BackendInitException` with inspectable per-backend attempts.
- The helper catches `Exception` only, so programming/runtime `Error`s are not swallowed.
- Fallback warnings now use `developer.log(level: 900, name: 'flutter_gemma')`.
- `activeBackend` docs now call out silent fallback behavior.
- The new aggregate exception types are exported from the public package API.
- Added focused tests for fallback order, successful fallback, aggregate failure details, public export coverage, and non-caught programming errors.

Validation:
- `flutter test --no-pub`
- targeted `flutter analyze --no-pub`
- GitHub `analyze-and-test` is green

DenisovAV:
Thanks Natan. CI green, all points addressed. Will fold into
release/0.16.2.

Sasha

PR #304, opened 2026-06-03
Show downloaded models list

fotiDim:
Making the example app a bit more intuitive by:
- Showing a list of downloaded models
- Showing a dialog for gated models that need to be granted permission through Huggingface
- Highlighting the currently loaded model
- Adding a lauch.json
- Allowing to easily copy the model path so that you can easily download and copy it in your own app
- Fixing a bug in the model download screen where the back arrow would not appear
- Adjusting back-navigation logic

Screenshot 2026-06-03 at 13 48 14
Screenshot 2026-06-03 at 12 09 21
Screenshot 2026-06-03 at 12 09 31

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,28 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "example",
+            "cwd": "example",
+            "request": "launch",
+            "type": "dart"
+        },
+        {
+            "name": "example (profile mode)",
+            "cwd": "example",
+            "request": "launch",
+            "type": "dart",
+            "flutterMode": "profile"
+        },
+        {
+            "name": "example (release mode)",
+            "cwd": "example",
+            "request": "launch",
+            "type": "dart",
+            "flutterMode": "release"
+        }
+    ]
+}
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,7 +1,4 @@
 {
-    // Use IntelliSense to learn about possible attributes.
-    // Hover to view descriptions of existing attributes.
-    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
     "version": "0.2.0",
     "configurations": [
         {
--- a/example/integration_test/active_model_restore_test.dart
+++ b/example/integration_test/active_model_restore_test.dart
@@ -18,7 +18,6 @@ import 'dart:io';
 import 'package:flutter/services.dart' show rootBundle;
 import 'package:flutter_gemma/flutter_gemma.dart';
 import 'package:flutter_gemma/core/di/service_registry.dart';
-import 'package:flutter_gemma/core/model.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
 import 'package:path_provider/path_provider.dart';
--- a/example/lib/chat_screen.dart
+++ b/example/lib/chat_screen.dart
@@ -3,7 +3,6 @@ import 'package:flutter_gemma/flutter_gemma.dart';
 import 'package:flutter_gemma_example/chat_widget.dart';
 import 'package:flutter_gemma_example/loading_widget.dart';
 import 'package:flutter_gemma_example/models/model.dart';
-import 'package:flutter_gemma_example/model_selection_screen.dart';
 import 'package:flutter_gemma_example/services/auth_token_service.dart';
 
 class ChatScreen extends StatefulWidget {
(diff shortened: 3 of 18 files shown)


DenisovAV (commented):
Thanks @fotiDim — nice UX work, CI's green. A few things in the new code before merge:

1. **`downloaded_model_deleter.dart`** — deleting an embedding model deletes the shared `sentencepiece.model` tokenizer (all embedding models share it, and `uninstallModel` has no refcounting), breaking the others. The confirm dialog also says "and its tokenizer". Only delete it if no other installed embedding model still uses it, and make the dialog copy conditional.
2. **`gated_model_access_dialog.dart`** — "Open on Hugging Face" silently does nothing if `canLaunchUrl` is false (or `launchUrl` throws), and that's the action that unblocks a gated download. Add a fallback (copy URL / SnackBar).
3. **`downloaded_model_deleter.dart`** — deleting the active model frees the native handle but doesn't clear the manager's active spec, so `hasActiveModel()` / `activeInferenceModelId()` stay stale afterwards.

Minor: `installed_model_lookup.dart` reaches into internal APIs (`core/di/service_registry.dart`, `core/domain/model_source.dart`) for path resolution — that's on us to expose publicly, not a blocker.

Heads-up: the web gated detection relies on typed 401/403 that our web download path doesn't emit yet (it collapses them into `DownloadError.unknown`) — pre-existing in our `lib/`, untouched by this PR. I'll fix it separately.

Thanks!

fotiDim:
@DenisovAV see my last commit. For a cleaner solution I had to add a couple of APIs.

DenisovAV:
Thanks @fotiDim — this is a genuinely nice contribution. The downloaded-models list, gated-model dialog, delete/copy-path, and active-model highlight make the example app much more useful, and the `clearActiveInferenceIdentity`/`clearActiveEmbeddingIdentity` API + the `DownloadError`/`DownloadException` export fill real gaps. The mobile/web impls are symmetric, PreferencesKeys are used correctly (no inline strings), and resource cleanup (closing models before clearing identity) is sound.

I ran a full multi-agent review. A few things to address before merge:

### Must fix

**1. `debugPrint` in `lib/` → use `gemmaLog`.** The new `clearActive*Identity` methods add 8 `debugPrint` calls in `mobile_model_manager.dart` and `web_model_manager.dart`. 0.16.5 migrated all `lib/` logging to `gemmaLog()` (silent in release, level-gated) and bans raw `debugPrint` in `lib/`. Your branch predates that migration, so it'll surface on rebase — please switch all 8 to `gemmaLog(...)` (the adjacent `clearModelCache` is now `gemmaLog('Model cache cleared')`). `flutter analyze` won't flag this, so it's easy to miss. (The `debugPrint` in `example/` is fine — the rule is `lib/`-only.)

**2. The exported `DownloadException` doesn't actually reach catch sites on mobile.** The whole point of exporting it is `} on DownloadException catch (e) { switch (e.error) … }` — but on mobile, `mobile_model_manager.dart` (~line 366) catches the `DownloadException` and re-wraps it as `ModelDownloadException` (the original becomes a `dynamic` `.cause`). So app code can't type-match it, and `ModelDownloadException`/`ModelException` aren't exported either. Your own `gated_model_access_dialog.dart` shows the symptom — it can't type-match and falls back to `toString().contains('401'/'403')` substring matching. Either stop re-wrapping `DownloadException` on the mobile path so the sealed type propagates intact, or export the wrapper with an `Object?` cause. Right now the typed-error feature is decorative on mobile.

### Should fix

**3. `clearActive*Identity` can silently leave dangling prefs.** Both methods null the in-memory spec *before* the `try { prefs.remove(...) } catch (e) { log }`, then report success unconditionally. If `prefs.remove` throws, the in-memory state is cleared but the persisted keys survive — and on next launch the restore path reads an active model whose file was just deleted. Please clear the in-memory field only *after* a successful `prefs.remove`, and surface/rethrow the failure instead of swallowing it (we avoid silent fallbacks).

**4. Adding abstract methods to `ModelFileManager` is a breaking change** for any third-party implementer (it's a public exported type). Please add a one-line breaking note in CHANGELOG.md — and ideally give the new methods a default implementation so it's non-breaking.

**5. Consider clearing active identity inside `uninstallModel()` itself** rather than as a separate verb the caller must remember to pair after uninstall — that way every consumer benefits and prefs can't dangle. Right now the pairing lives only in the example deleter.

### Minor
- The 8 active-model prefs keys are now hand-maintained in 3 places per manager (persist/restore/clear) × 2 managers — a shared helper owning the key list would prevent drift (it's already slightly drifted: persist writes `*Source` keys restore never reads).
- `downloaded_model_loader.dart` `load()` dispatches on the `sealed DownloadedCatalogMatch` with an `if`-chain — a `switch` would make it exhaustive (a future match subtype currently no-ops silently).
- `installed_model_lookup.dart` reaches into non-exported `core/` internals (`ServiceRegistry`, `fileSystemService.getReadTargetPath`) — resolving an installed model's on-disk path is generally useful and would be better as a public `FlutterGemma` API than reconstructed in `example/`.
- The gated-error classifier requires both a 401/403 token *and* a "gated"-flavored word — a terse "HTTP 403" slips through to the generic error. A 401/403 status alone should be enough to show the access dialog.

Really appreciate the work here — once 1–4 are addressed I'll re-review. Happy to help with the `DownloadException` propagation question if useful.

DenisovAV:
Thanks @fotiDim — shipped your downloaded-models feature in **1.0.3** 🎉 The branch couldn't merge as-is (it predated the 1.0 `packages/` monorepo split → conflicts), so I ported it onto the new structure with all the review fixes in #338, with you credited as co-author. Closing in favor of #338. Really appreciate the contribution! 🙏

fotiDim:
@DenisovAV Cool. I was planning to work on it today but you were faster 🚀

Issue #170, opened 2026-01-15
The issue is that after testing a faulty model, reverting to the old model results in it not working.


gray-seven:
1. Running a faulty model, for example, results in the error below (e.g., FastVLM-0.5B.litertlm on iPhone 11).
2. Switching to a working model, the same error persists when trying to chat.
3. Closing and reopening the app allows it to work.

```
flutter: ℹ️  Model already installed: FastVLM-0.5B.litertlm (skipping download)
flutter: ✅ Set active inference model: FastVLM-0.5B
flutter: ✅ Inference model installed and set as active: FastVLM-0.5B
flutter: [GemmaChatService] Step 1: Model installed ✅
flutter: [GemmaChatService] Step 2: Creating InferenceModel...
2
flutter: UnifiedModelManager: Model FastVLM-0.5B installed: true
flutter: Using unified model file: /var/mobile/Containers/Data/Application/24C87C64-2FA0-4823-952E-1AAFA3C2E6CC/Documents/FastVLM-0.5B.litertlm
flutter: [GemmaChatService] getActiveModel attempt 1/3 failed: PlatformException(failedToInitializeEngine(Optional("INTERNAL: RET_CHECK failure (third_party/odml/infra/genai/inference/utils/llm_utils/model_data.cc:424) model Error building tflite model.\n=== Source Location Trace: === \nthird_party/odml/infra/genai/inference/utils/llm_utils/model_data.cc:424\nthird_party/odml/infra/genai/inference/utils/llm_utils/config_utils.cc:558\nthird_party/odml/infra/genai/inference/llm_engine.cc:2132\n")), GenAiInferenceError, Stacktrace: ["0   Runner.debug.dylib                  0x00000001067449ac $s13flutter_gemma9wrapError33_AC1DF4B16F5474DCF2184B6087243948LLySayypSgGypF + 1832", "1   Runner.debug.dylib                  0x000000010674af38 $s13flutter_gemma20PlatformServiceSetupC5setUp15binaryMessenger3api20messageChannelSuffixySo013FlutterBinaryI0_p_AA0cD0_pSgSStFZyypSg_yAKctcfU_ys6ResultOyyts5Error_pGcfU_ + 324", "2   Runner.debug.dylib                  0x000000010672d08c $s13flutter_gemma19PlatformService<…>
[...]

PR #309, opened 2026-06-07
fix(mobile): give each createSession a fresh native session (#308)

### LINKED ISSUE #308: MobileInferenceModel.createSession short-circuits on cached _createCompleter, causing createChat to share native sessions across calls
## Observed

After the first call to `createSession` succeeds, `_createCompleter` is cached and never nulled until the resulting `MobileInferenceModelSession.close()` fires its `onClose` callback. Every subsequent `createSession` call (including those reached via `createChat()` → `initSession()` → `sessionCreator()`) returns the same cached future, so a new `InferenceChat` wrapper ends up backed by the same native session — with all prior turn tokens still primed in the LiteRT KV cache.

Reference: [`flutter_gemma_mobile_inference_model.dart`](https://github.com/DenisovAV/flutter_gemma/blob/main/lib/mobile/flutter_gemma_mobile_inference_model.dart) (version checked: `0.15.2`):

```dart
if (_createCompleter case Completer<InferenceModelSession> completer) {
  return completer.future;     // <-- returns the same cached session
}
final completer = _createCompleter = Completer<InferenceModelSession>();
...
final session = _session = MobileInferenceModelSession(
  ...
  onClose: () {
    _session = null;
    _createCompleter = null;     // <-- only path that clears the cache
  },
);
```

## Expected

`createChat()` should either:

(a) provision a fresh native session (matching the Dart-side `InferenceChat` lifecycle — each `createChat` returns a new `InferenceChat`, and a reasonable reading of the API is that each chat gets its own session), **or**

(b) document the cached behaviour and require callers to `await chat.session.close()` before calling `createChat()` again.

## Why this matters

A single `InferenceModel` driving multiple unrelated chats (e.g. an app with two distinct LLM-driven channels) sees cross-channel context bleed. The application appears to send a clean preamble; the model conditions on the prior chat's KV cache anyway. Logs that capture the bytes sent to `addQueryChunk` show nothing wrong — the bleed is invisible without inspecting the LiteRT engine's internal state.

In our case the symptom was a Companion chat reply referencing a recent PvP debrief match the Companion's prompt never mentioned. We discovered the root cause by reading the plugin source after the bug was unreproducible in any prompt-injection test.

## Minimal repro (sketch)

```dart
final model = await FlutterGemma.instance.createModel(
  modelType: ModelType.gemmaIt,
  maxTokens: 2048,
);

final chatA = await model.createChat(temperature: 0.8);
await chatA.addQueryChunk(
  Message.text(text: 'Remember the secret X.', isUser: true),
);
final _ = await chatA.generateChatResponse();

// Build chat B without closing chat A:
final chatB = await model.createChat(temperature: 0.8);
await chatB.addQueryChunk(
  Message.text(text: 'What secret do you know?', isUser: true),
);
final response = await chatB.generateChatResponse();
// Observation: response references X, even though chatB's Dart-side
// history is empty.
```

`chatA.session` and `chatB.session` are the **same** `MobileInferenceModelSession` instance.

## Workaround

Callers explicitly `await chatA.session.close()` before `createChat()` for chat B. This fires the `onClose` callback that nulls `_session` / `_createCompleter`, so the next `createSession` provisions a genuinely fresh native session with an empty KV cache.

We've implemented this discipline at our `LlmService` boundary — closing the previous chat's session before every `createChat` call, with an awaitable handshake for fire-and-forget reset paths so a quick reset → new turn sequence doesn't race the close.

## Suggestion

Either:

(a) reset `_createCompleter` after the cached session is consumed once so subsequent `createSession` calls allocate fresh native sessions (probably also nulling `_session` so the field reflects "no active session"), **or**

(b) add a doc comment on `createChat` (and `createSession`) calling out the required explicit close discipline. Even just documenting it would protect future callers — discovering the cached behaviour by KV-cache bleed in production is a rough debugging experience.

Happy to send a PR if either direction is preferred. Thanks for the plugin!

mtmanty:
## Description

Fixes #308.

`MobileInferenceModel.createSession` cached its `_createCompleter` and only cleared it from the session's `onClose`, so the completer acted as a **permanent cache** rather than an in-flight guard. Every `createChat` after the first short-circuited on the cached completer and returned the *same* `MobileInferenceModelSession` wrapper — and crucially never re-invoked the native `createSession`, which is the call that closes the prior session and opens a fresh one:

```kotlin
// FlutterGemmaPlugin.createSession (Android)
session?.close()
session = currentEngine.createSession(config)
```

So a single `InferenceModel` driving two sequential chats shared one native session, and the first conversation's KV cache bled into the second. The app sends a clean prompt; the model still conditions on the previous context. Because `LlmPromptLog`-style logging captures the bytes sent to `addQueryChunk` (not the engine's KV cache), the bleed is invisible without inspecting native state — which is what made the original report hard to pin down.

### The fix

The completer is now a **pure in-flight guard**, cleared in a `finally` once creation settles (success *or* failure):

- **Sequential** `createSession` calls fall through to the native create-and-replace, so each chat gets a clean KV cache (the issue #308 fix).
- A **failed** creation no longer caches a rejected future that blocks retry — the same bug class already fixed at the model level in #170 (`model_creation_failure_test.dart`).
- **Concurrent** callers still dedupe onto one in-flight creation (the guard's original purpose is preserved).

Supporting changes for the singleton-session lifecycle:

- `createSession` closes the prior `_session` before creating the next, so the orphaned wrapper's Dart-side resources (event subscription, stream controller) are released and stray calls on it throw `Model is closed` cleanly instead of silently hitting the new native session.
- `onClose` is identity-guarded so a late close of a superseded session can't null a newer `_session`, and it no longer touches `_createCompleter` (owned by the `finally` now).
- `MobileInferenceModelSession.close()` is now **idempotent** — a second close (e.g. a consumer closing a superseded chat) must not run native `closeSession()` again, which would tear down the *current* session (`closeSession` is argument-less).
- `createSession` returns `completer.future` so the caller stays the error listener after the completer field is cleared (mirrors the `createModel` idiom from #170).

### Scope note

The FFI/litert legacy `createSession` (`lib/core/ffi/ffi_inference_model.dart`) has the same completer-cache shape. I left it out to keep this PR scoped to the issue (and the MediaPipe path is what #308 names) — happy to follow up with the same treatment if you'd like it in here or as a separate PR.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring
- [x] Test addition

## Testing
- [ ] Tested on Android
- [ ] Tested on iOS
- [ ] Tested on Web
- [ ] Tested on Desktop (macOS/Windows)

Logic is covered by a new unit test rather than device testing — see below. The full existing suite (`flutter test`) passes (393 tests) and `flutter analyze` is clean on the changed files.

`test/mobile/session_creation_reuse_test.dart` is a distilled Buggy/Fixed logic pair mirroring `model_creation_failure_test.dart` (#170): it models the completer + session lifecycle (no platform channel needed) and asserts the bug (shared session across `createChat`, blocked retry after failure) and the fix (fresh native session per chat, prior session closed first, retry after failure works, concurrent dedupe preserved, idempotent close doesn't tear down the live session).

## Checklist
- [x] Self-review completed
- [x] Comments added for complex code
- [ ] Documentation updated (no public API change)
- [x] No new warnings (`flutter analyze` clean on changed files)
- [x] Tests added/updated

### DIFF
--- a/lib/mobile/flutter_gemma_mobile.dart
+++ b/lib/mobile/flutter_gemma_mobile.dart
@@ -239,6 +239,14 @@ class MobileInferenceModelSession extends InferenceModelSession {
 
   @override
   Future<void> close() async {
+    // Idempotent: a second close must not run native `closeSession()`
+    // again. With the singleton-session lifecycle (a fresh createSession
+    // supersedes the prior wrapper), a superseded wrapper and the live
+    // one can both be closed by a caller; without this guard the stale
+    // close would tear down the *current* native session (closeSession
+    // is argument-less and closes whatever session is active). See the
+    // createSession comment and issue #308.
+    if (_isClosed) return;
     _isClosed = true;
 
     // Cancel event subscription first to stop receiving events
--- a/lib/mobile/flutter_gemma_mobile_inference_model.dart
+++ b/lib/mobile/flutter_gemma_mobile_inference_model.dart
@@ -127,11 +127,35 @@ class MobileInferenceModel extends InferenceModel {
       throw StateError(
           'Model is closed. Create a new instance to use it again');
     }
+    // Single-flight guard for genuinely *concurrent* callers only. Unlike
+    // the model singleton, a session is NOT reused across calls: each
+    // createSession (and therefore each createChat) must yield a fresh
+    // native session with a clean KV cache. The completer is cleared in
+    // the `finally` below once creation settles, so a *sequential* second
+    // call falls through to the native createSession — which closes the
+    // prior session and creates a new one (FlutterGemmaPlugin.createSession
+    // does `session?.close(); session = engine.createSession(...)`) —
+    // instead of returning the stale wrapper. Without this, the cached
+    // completer made every later createChat reuse the first session, so
+    // the previous conversation's KV cache bled into the next chat (the
+    // app sends a clean prompt; the model still conditions on the old
+    // context). See https://github.com/DenisovAV/flutter_gemma/issues/308.
     if (_createCompleter case Completer<InferenceModelSession> completer) {
       return completer.future;
     }
     final completer = _createCompleter = Completer<InferenceModelSession>();
     try {
+      // Close any prior singleton session before creating the next so its
+      // Dart-side resources (event subscription, stream controller) are
+      // released and stray calls on the old wrapper throw `Model is
+      // closed` cleanly instead of silently hitting the new native
+      // session. The native layer also closes the old session, but doing
+      // it here keeps the orphaned wrapper's `_isClosed` flag honest and
+      // means at most one live wrapper maps to the single native session.
+      if (_session case final previous?) {
+        await previous.close();
+      }
+
       // LoRA support is fully integrated via Modern API (InferenceInstallationBuilder)
       final resolvedLoraPath = loraPath;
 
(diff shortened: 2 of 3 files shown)


DenisovAV:
Thanks @mtmanty — and thanks for surfacing the bug. You're right: the original assumption was that the previous session always gets closed before a new one is created (which clears the cached completer via `onClose`). But you're correct that the plugin should close it itself when the caller doesn't — which is exactly what your fix does with `await previous.close()`.

Merging, ships in the next release. Yes please to a separate PR for the FFI/litert path.

Thanks again!

mtmanty:
Thanks for the quick merge! Opened the FFI/litert follow-up as requested: #310 — same `createSession` cached-completer fix (in-flight guard cleared on settle, close prior session before opening a fresh handle, idempotent `close()`), with a `session_creation_reuse_test.dart` mirroring the one here. One difference noted in the PR: the FFI path already cleared the completer on failure, so cache-on-success was its only bug.

PR #310, opened 2026-06-07
fix(ffi): give each createSession a fresh conversation handle (#308)

mtmanty:
## Description

Follow-up to #309 (as requested in that thread) — the same `createSession` cached-completer fix for the **litert/FFI** path.

`FfiInferenceModel.createSession` cached its `_createCompleter` and only cleared it from the session's `onClose`, so on **success** the completer acted as a permanent cache: every `createChat` after the first short-circuited on the cached completer and returned the *same* `FfiInferenceModelSession` — never reaching the `await _session?.close()` + fresh `createConversationHandle()` lines below. The result is the same KV-cache bleed #309 fixed for MediaPipe: the previous conversation's context leaks into the next chat.

> Note: unlike the MediaPipe path, the FFI path already cleared the completer in its `catch`, so a *failed* `createSession` did **not** block retry here — the cache-on-success bleed was the only FFI bug.

### The fix (mirrors #309)

- The completer is now a pure in-flight guard, cleared in a `finally` once creation settles, so a sequential second call opens a fresh conversation handle (closing the prior session first).
- `onClose` is identity-guarded so a late close of a superseded session can't null a newer `_session`, and no longer touches `_createCompleter`.
- `FfiInferenceModelSession.close()` is now idempotent — a second close (a consumer closing a superseded session; `createSession` also closes the prior one) must not re-run `handle.close()` / `onClose()`.
- `createSession` returns `completer.future` so the caller stays the error listener after the completer field is cleared.

## Type of Change
- [x] Bug fix
- [x] Test addition

## Testing
- [ ] Tested on Android
- [ ] Tested on iOS
- [ ] Tested on Web
- [ ] Tested on Desktop (macOS/Windows)

Logic is covered by a unit test rather than device testing. `test/core/ffi/session_creation_reuse_test.dart` is a distilled Buggy/Fixed logic pair mirroring `test/mobile/session_creation_reuse_test.dart` from #309 (the FFI client's `createConversationHandle` returns a concrete handle type that can't be faked through the public surface, so the same distilled approach is used). It asserts the cache-on-success bleed (shared handle across `createChat`) and the fix (fresh handle per chat, prior session closed first, retry after failure still works, concurrent dedupe preserved, idempotent close). The full `test/core/ffi` suite passes (25 tests); `flutter analyze` is clean on the changed files.

## Checklist
- [x] Self-review completed
- [x] Comments added for complex code
- [ ] Documentation updated (no public API change)
- [x] No new warnings
- [x] Tests added/updated

### DIFF
--- a/lib/core/ffi/ffi_inference_model.dart
+++ b/lib/core/ffi/ffi_inference_model.dart
@@ -101,6 +101,17 @@ class FfiInferenceModel extends InferenceModel {
     final sessionSw = Stopwatch()..start();
 
     try {
+      // Legacy singleton lane: close the previous conversation BEFORE
+      // opening a fresh one. The engine holds at most one live
+      // conversation (upstream litert-lm #966), so closing first never
+      // leaves two alive — matching the delete-before-create order the
+      // virtual-session multiplexer already uses. It also means a
+      // teardown error can't leak a handle that was created first and
+      // then left unwrapped: each session owns its own conversation
+      // pointer, so closing the old one can't touch the not-yet-created
+      // new one. See PR #310 review.
+      await _session?.close();
+
       // For Gemma 4, push tools into the SDK conversation config so it can
       // render native `<|tool>declaration:...<tool|>` tokens via minja. Other
       // model types still use Dart-side prompt injection in chat.dart.
@@ -109,8 +120,6 @@ class FfiInferenceModel extends InferenceModel {
           : null;
 
       final beforeConv = sessionSw.elapsedMilliseconds;
-      // Legacy singleton lane: close the previous conversation (if any) and
-      // open a fresh one. Mirrors the pre-multi-session overwrite contract.
       final handle = ffiClient.createConversationHandle(
         systemMessage: systemInstruction,
         toolsJson: toolsJson,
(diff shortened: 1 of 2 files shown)


DenisovAV:
Thanks @mtmanty — the lifecycle fix is correct. One thing to change before merge: the ordering.

`createSession` currently creates the new handle first, then closes the prior session:

```dart
final handle = ffiClient.createConversationHandle(...);
await _session?.close();
```

Could you flip it to close-first?

```dart
await _session?.close();
final handle = ffiClient.createConversationHandle(...);
```

Two reasons:
- The engine allows only one live conversation at a time (upstream #966) — create-first briefly has two live, which the rest of the FFI layer deliberately avoids (the virtual-session multiplexer always deletes-before-creates). The in-code comment already says "close the previous … and open a fresh one", which is the close-first order.
- If `_session?.close()` throws after the new handle is created, the new handle leaks (it's never wrapped or closed) and the teardown error surfaces as a create error. Close-first avoids both — each session owns its own conversation pointer, so closing the old one can't touch the not-yet-created new one.

Worth updating the test's `Fixed` model to the same order so it documents the intent. Thanks!

mtmanty:
Done in 2a6e9ef — flipped to close-first (`await _session?.close()` then `createConversationHandle(...)`), so there is never more than one live conversation and a teardown error can no longer leak the new handle. Updated the test's `Fixed` model to the same order to document the intent. Thanks for the review!

DenisovAV:
Thanks @mtmanty!

PR #312, opened 2026-06-08
fix: neutralize U+FFFD in model-output logs to prevent flutter run crash (#306)

### LINKED ISSUE #306: debugPrint unable to handle UTF-8 replacement characters coming from the model
If the model responds with a replacement character any call to debugPrint with a string containing the replcaemnet character crashes the entire app

Easily reproducible by asking Gemma4 "What is the UTF-8 replacement character?"

Failure comes from this line when called on Web: https://github.com/DenisovAV/flutter_gemma/blob/main/lib/web/flutter_gemma_web.dart#L864
But any call to debugPrint would break on this behavior

https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/convert.dart#L81

Since the flutter SDK debugPrint call doesn't support the encoding, I imagine a safe solution would be to create a wrapper around debugPrint that replaces � with some string that can be safely rendered, like `U+FFFD` or something similar?

Debug console from crash: 
```
--- Sending to Native ---
History:

Current Message:
What is the UTF-8 replacement character?
-------------------------
🟢 WebModelSession.addQueryChunk() called - hasImage: false, hasAudio: false, supportImage: false, supportAudio: false
[transformToChatPrompt] modelType=ModelType.gemma4, fileType=ModelFileType.task, messageType=MessageType.text, isUser=true
[transformToChatPrompt] Using _formatToolResponseContent, result length=40
🟢 Added text part: What is the UTF-8 replacement character?...
🟢 Total prompt parts: 1
InferenceChat: Starting async stream generation
InferenceChat: Starting to iterate over native tokens...
🌊 getResponseAsync: Starting async response generation
🔧 _createPromptArray: Starting with 1 prompt parts
📝 _createPromptArray: All text parts, returning string of length 40
📝 _createPromptArray: Text preview: What is the UTF-8 replacement character?...
🎯 getResponseAsync: Prompt array type: String
🎯 getResponseAsync: Is JSString? true
📝 getResponseAsync: Using generateResponse for text-only prompt
📝 getResponseAsync: Received partial (complete: false): 

The UTF...
InferenceChat: Received filtered token: "

The UTF"
InferenceChat: No function processing, emitting token as text: "

The UTF"
📝 getResponseAsync: Received partial (complete: false): -8 replacement...
InferenceChat: Received filtered token: "-8 replacement"
InferenceChat: No function processing, emitting token as text: "-8 replacement"
📝 getResponseAsync: Received partial (complete: false):  character is **...
InferenceChat: Received filtered token: " character is **"
InferenceChat: No function processing, emitting token as text: " character is **"
Bad UTF-8 encoding (U+FFFD; REPLACEMENT CHARACTER) found while decoding string: 📝 getResponseAsync: Received partial (complete: false): �** (...
. The Flutter team would greatly appreciate if you could file a bug explaining exactly what you were doing when this happened:
https://github.com/flutter/flutter/issues/new/choose
The source bytes were:
[240, 159, 147, 157, 32, 103, 101, 116, 82, 101, 115, 112, 111, 110, 115, 101, 65, 115, 121, 110, 99, 58, 32, 82, 101, 99, 101, 105, 118, 101, 100, 32, 112, 97, 114, 116, 105, 97, 108, 32, 40, 99, 111, 109, 112, 108, 101, 116, 101, 58, 32, 102, 97, 108, 115, 101, 41, 58, 32, 239, 191, 189, 42, 42, 32, 40, 46, 46, 46, 10]
3

Exited (1).
```

Relevant XKCD: https://xkcd.com/1913/

DenisovAV:
## Description

Fixes #306.

Model output can legitimately contain the Unicode replacement character (U+FFFD, `�`) — e.g. when the model is asked about the replacement character itself (the issue's repro: *"What is the UTF-8 replacement character?"*), or when it emits a byte-fallback token. Logging that text via `debugPrint` sends `�` to the app's stdout, where `flutter run`'s tool-side reader treats it as a malformed-output signal and calls `throwToolExit`, **killing the dev session** (flutter_tools/lib/src/convert.dart). The app itself is fine — this only crashes `flutter run` at dev time.

### Fix

New `lib/core/utils/safe_debug_print.dart`:
- `safeDebugPrint(String?)` — drop-in `debugPrint` replacement that rewrites `�` → literal `U+FFFD` before printing.
- `sanitizeForLog(String)` — same rewrite for call sites that build a log string by hand (e.g. truncating helpers).

The replacement keeps the log visible (and still shows that a replacement char was present) rather than dropping it — as suggested in the issue.

Applied **only at log sites that print model-produced text**, not globally (ordinary plugin log lines never contain U+FFFD):
- `lib/web/flutter_gemma_web.dart` — response preview + text/multimodal partials (the crash site from the repro)
- `lib/core/chat.dart` — filtered token, emitting token, accumulated response
- `lib/core/ffi/litert_lm_client.dart` — native-log chunk dump
- `lib/core/image_error_handler.dart` — prompt/response logging sanitizers

## Type of Change
- [x] Bug fix
- [x] Test addition

## Testing
`test/core/utils/safe_debug_print_test.dart` (9 tests, all pass) covers: U+FFFD replacement, multiple occurrences, ordinary/emoji text untouched, the exact bytes from the #306 repro, and `safeDebugPrint` forwarding (sanitized / unchanged / null). `flutter analyze` clean; `flutter build web` succeeds.

## Checklist
- [x] Self-review completed
- [x] Tests added
- [x] No new warnings (`flutter analyze` clean)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,28 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "example",
+            "cwd": "example",
+            "request": "launch",
+            "type": "dart"
+        },
+        {
+            "name": "example (profile mode)",
+            "cwd": "example",
+            "request": "launch",
+            "type": "dart",
+            "flutterMode": "profile"
+        },
+        {
+            "name": "example (release mode)",
+            "cwd": "example",
+            "request": "launch",
+            "type": "dart",
+            "flutterMode": "release"
+        }
+    ]
+}
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,7 +1,4 @@
 {
-    // Use IntelliSense to learn about possible attributes.
-    // Hover to view descriptions of existing attributes.
-    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
     "version": "0.2.0",
     "configurations": [
         {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.16.5
+- **Fix `debugPrint` crash on U+FFFD in model output** (#306, thanks @pchang0010): `safeDebugPrint` neutralizes `�` so it can't kill `flutter run`.
+- **Fix session KV-cache bleed across `createChat`** (#308, thanks @mtmanty): each MediaPipe session gets a fresh native session.
+
 ## 0.16.4
 - **Fix embedding freezing the UI thread** (#299): forward pass runs on a background isolate.
 - **Fix Windows build on non-UTF-8 locales** (#212): add `/utf-8` to the MSVC plugin target.
--- a/example/integration_test/active_model_restore_test.dart
+++ b/example/integration_test/active_model_restore_test.dart
@@ -18,7 +18,6 @@ import 'dart:io';
 import 'package:flutter/services.dart' show rootBundle;
 import 'package:flutter_gemma/flutter_gemma.dart';
 import 'package:flutter_gemma/core/di/service_registry.dart';
-import 'package:flutter_gemma/core/model.dart';
 import 'package:flutter_test/flutter_test.dart';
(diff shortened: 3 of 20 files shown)


DenisovAV:
Closing — branch was accidentally based on an unmerged branch (#304), polluting the diff. Reopening from a clean main base.

Issue #316, opened 2026-06-12
flutter_gemma native-v0.12.0 litertlm-android_arm64.tar.gz checksum mismatch on Android build


remingtonc:
Encountering an issue with the build hook downloads for Android specifically.
```
flutter_gemma: Downloading litertlm native libs from https://github.com/DenisovAV/flutter_gemma/releases/download/native-v0.12.0/litertlm-android_arm64.tar.gz ...

flutter_gemma: Checksum mismatch for litertlm-android_arm64.tar.gz!

  Expected: e24804d922aadd91a85a6faf272a20e9c3e7991ed2754cf2a9071ad08a8fc2ce
  Actual:   f809c5a29867062cda74186c7ebebd500a2f36d0b6ad6f7ca8eab902af7fc784

flutter_gemma: Downloading qdrant_edge native libs from https://github.com/DenisovAV/flutter_gemma/releases/download/qdrant-edge-v0.6.1-flutter1/qdrant-edge-android_arm64.tar.gz ...

flutter_gemma: Checksum verified (qdrant-edge-android_arm64.tar.gz)

flutter_gemma: qdrant_edge libs cached to /root/.cache/flutter_gemma/native/qdrant_edge/android_arm64
```

Been working fine generally until now. Run `flutter clean` and `flutter pub cache clean` several times. Deleted every flutter related `/root/.some_dir` directory I could imagine to try and ensure not a cache issue.

Going to destroy the container to re-verify, but somehow it's stuck. Opening this issue to log, as someone will eventually encounter this checksum issue similarly, and I don't see one in issues tracker.

remingtonc:
Ah, it looks like `litertlm-android_arm64.tar.gz` was updated about 2 weeks ago, while everything else was uploaded 3 weeks ago. But the checksum in `checksums_litertlm.txt` was not updated as well (assuming that's where this is deriving).

remingtonc:
@DenisovAV I am not inclined to bypass the checksum check until you confirm it was an intentional change in the release. 🙂
There doesn't seem to be a changelog of sorts for a Release's Assets.

DenisovAV:
Hi @remingtonc — thanks for logging this so clearly, and for **not bypassing the checksum** — that's exactly the right call, the check did its job.

[...]

PR #321, opened 2026-06-16
docs(release-skill): port #316 three-way checksum-consistency check

DenisovAV:
Carries the mandatory checksum-consistency rule that landed on the monolith in #317 but didn't reach the 1.0 split's release SKILL.md. Adapts the hook path to the monorepo (`packages/flutter_gemma_litertlm/hook/build.dart`).

#316: a stale `checksums_litertlm.txt` (asset `f809c5a2` vs txt `e24804d9`) sent a user down the wrong debug path. The rule: the same SHA must match in all three places — the uploaded `.tar.gz`, `checksums_litertlm.txt` on the Release, and the hook's `_checksums` map.

Docs-only change to local release tooling; no shipped code touched.

### DIFF
--- a/.claude/skills/release/SKILL.md
+++ b/.claude/skills/release/SKILL.md
@@ -180,6 +180,40 @@ for f in "$DIST"/litertlm-*.tar.gz; do
 done
 ```
 
+### ⛔ Three-way checksum consistency — MANDATORY (regression: #316)
+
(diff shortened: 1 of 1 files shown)


DenisovAV:
Superseded by the release/1.0.0 PR — the #316 checksum-consistency note is carried there alongside the rest of the 1.0.0 release work (qdrant 16KB fix, version bumps).

PR #342, opened 2026-06-24
feat(core): injectable shared FileDownloader updates stream at init

hughesyadaddy:
## Summary

Adds optional `downloadUpdatesStream: Stream<TaskUpdate>?` and
`fileSystemService` parameters to `FlutterGemma.initialize()`, wired
through `ServiceRegistry` into `SmartDownloader`.

## Problem

`FileDownloader().updates` is **single-subscription**. Many host apps
(including production Flutter apps using `background_downloader` for both
media cache and on-device LLM installs) already forward that stream
through their own **broadcast hub**.

`SmartDownloader` also wraps `FileDownloader().updates` with its own
`asBroadcastStream()`. Whichever wrapper subscribes first wins; the loser
can miss `TaskProgressUpdate` events. In practice this shows up as **AI
model installs stuck at 0%** on a shared progress UI, especially when
**attaching to an in-progress download** after restart (the attach path
always used the internal stream).

## Solution

```dart
await FlutterGemma.initialize(
  downloadUpdatesStream: hostBroadcastStream, // Stream<TaskUpdate>?
  inferenceEngines: [MediaPipeEngine(), LiteRtLmEngine()],
);
```

`SmartDownloader` resolves streams in order:

1. Per-call override (if any)
2. Init-time `downloadUpdatesStream`
3. Internal singleton broadcast (unchanged default)

Both **new downloads** and **attach-to-existing** paths use the resolver.

`fileSystemService` is also threaded through `FlutterGemma.initialize()`
so apps with custom model storage (outside Documents) can still use the
official init entrypoint and engine registration — no need to call
`ServiceRegistry.initialize()` directly.

## Test plan

- [x] Unit test for configure/reset lifecycle (`download_updates_stream_config_test.dart`)
- [x] Manual: two consumers (cache client + Gemma install) share one broadcast hub; progress reaches 100%
- [x] Manual: attach-to-existing download after app restart receives progress events

## Backward compatibility

Fully backward compatible — defaults are `null`; existing behavior unchanged when not set.

### DIFF
--- a/packages/flutter_gemma/CHANGELOG.md
+++ b/packages/flutter_gemma/CHANGELOG.md
@@ -1,7 +1,5 @@
-## 1.1.1
-- Declare `libvndksupport.so` for the Android GPU backend — fixes Mali GPU hard-freeze on Android 12+ (#324).
-
 ## 1.1.0
+- Add optional `downloadUpdatesStream` and `fileSystemService` to `FlutterGemma.initialize()` for host apps that share a `background_downloader` broadcast hub or custom model storage paths (mobile only for the stream).
 - Add declared-column `Filter` support: `FilterSchema`/`FilterField`/`FilterFieldType` + `configure(FilterSchema)` on `VectorStoreRepository`.
 - Add optional `filterSchema:` to `FlutterGemma.initialize` (threaded to the vector store at registration).
 - Deprecate `enableHnsw` (no-op; vector search now runs inside the store engine).
--- a/packages/flutter_gemma/lib/core/api/flutter_gemma.dart
+++ b/packages/flutter_gemma/lib/core/api/flutter_gemma.dart
@@ -2,6 +2,7 @@ import 'package:flutter/foundation.dart' show kIsWeb;
 import 'package:flutter_gemma/core/api/inference_installation_builder.dart';
 import 'package:flutter_gemma/core/api/embedding_installation_builder.dart';
 import 'package:flutter_gemma/core/di/service_registry.dart';
+import 'package:flutter_gemma/core/services/file_system_service.dart';
 import 'package:flutter_gemma/core/domain/model_source.dart';
 import 'package:flutter_gemma/core/domain/web_storage_mode.dart';
 import 'package:flutter_gemma/core/infrastructure/web_download_service_stub.dart'
@@ -146,6 +147,20 @@ class FlutterGemma {
     // empty default keeps every store in its existing "filters are a safe
     // no-op" mode.
     FilterSchema filterSchema = const FilterSchema(),
+
+    /// Optional host-provided broadcast of download task updates (mobile).
+    ///
+    /// On iOS/Android, events must be [TaskUpdate] values from
+    /// `package:background_downloader` — the same shape as
+    /// [FileDownloader.updates]. Ignored on web. When omitted,
+    /// [SmartDownloader] uses its internal broadcast wrapper.
+    Stream<Object>? downloadUpdatesStream,
+
+    /// Optional custom model file storage (e.g. outside Documents).
+    ///
+    /// [FileSystemService] is defined in core but not exported from the
+    /// public barrel; pass an implementation from your app or tests.
+    FileSystemService? fileSystemService,
   }) async {
     // Migration: enableWebCache takes precedence if provided (for backward compatibility)
     final effectiveStorageMode = enableWebCache != null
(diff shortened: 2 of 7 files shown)


DenisovAV:
Thanks for this @hughesyadaddy — the problem is real and the diagnosis is spot-on: `FileDownloader().updates` is single-subscription and a host broadcast hub racing our internal `asBroadcastStream()` is exactly why installs stick at 0%. The injection approach is the right shape. A few things to address before merge — happy to help with any of them.

**Blocking:**

1. **The injected stream isn't guaranteed broadcast.** `_resolveUpdatesStream()` returns the injected `source` as-is, and it's `.listen()`'d at two sites (`smart_downloader.dart:270`/`:359`). The class supports multiple concurrent downloads, so two `.listen()` on a single-subscription injected stream throws `Bad state: Stream has already been listened to` → caught by the retry loop → 10 retries → bogus `DownloadError.unknown`. That re-creates the silent-fail this PR fixes, just for a different input. Suggest wrapping defensively: `source.isBroadcast ? source : source.asBroadcastStream()` (memoized like `_getUpdatesStream`).

2. **Core/web layering.** `service_registry.dart` is deliberately platform-split — every mobile/web import is behind `if (dart.library.js_interop)`, and `background_downloader` reaches the graph only via the conditional `mobile_service_factory`. The new bare `import 'package:flutter_gemma/mobile/smart_downloader.dart'` + the unconditional `SmartDownloader.configureDownloadUpdatesStream(...)` call run on web too, where it's meaningless. Please gate it (kIsWeb / the existing platform-factory seam).

3. **No `onError`/`onDone` on the injected-stream listens.** Both `.listen()` calls supply only `onData`. If the host hub forwards an error or closes, the download `Completer` never completes → the install hangs forever with no error. Please forward `onError`→`completeError` and `onDone`→complete-or-error.

**Should fix:**

4. `reset()` / `dispose()` don't clear the static `_configuredDownloadUpdatesStream` (the reset is `@visibleForTesting`), so a stale stream leaks across re-init.
5. Dead params: `_resolveUpdatesStream({override})` is never called with an argument, and the `downloadUpdatesStream` param added to `_create()` is never read (the wiring is the static call in `initialize()`).
6. Public-API coupling: exposing `Stream<TaskUpdate>` puts background_downloader's *sealed* type into core's public `initialize()` contract, and `FileSystemService` is an internal type not exported from the barrel. Worth a maintainer decision — possibly a small flutter_gemma-owned update type instead.

**Minor:** the test (`expect(true, isTrue)`) doesn't assert behavior — a `StreamController.broadcast()` + two concurrent consumers + an error-injection case would lock in the fix; and the two new `initialize()` params need dartdoc + a CHANGELOG line.

Really appreciate the contribution — the core idea is good, this is mostly about making it safe for the general case (non-broadcast input, web, hub errors). Glad to pair on the broadcast-wrap + onError bits if useful.

hughesyadaddy:
Thanks for the thorough review — all blocking items addressed in 64fa4c6:

**Blocking**
1. **Non-broadcast injected streams** — `_resolveUpdatesStream()` now memoizes `source.isBroadcast ? source : source.asBroadcastStream()` so concurrent downloads can each `.listen()` without `Bad state`.
2. **Core/web layering** — `SmartDownloader.configure…` moved behind `download_hub/configure_download_updates_stream_mobile.dart` with a web stub; `service_registry` no longer imports `mobile/smart_downloader.dart` unconditionally.
3. **Hub error/close propagation** — both attach + new-download `.listen()` calls now forward `onError` → progress/completer `completeError` and `onDone` → `completeError` when the hub closes early.

**Should fix**
- `ServiceRegistry.reset()` and `dispose()` call `clearDownloadUpdatesStream()` (wired to `SmartDownloader.clearConfiguration()`).
- Removed dead `downloadUpdatesStream` param from `_create()` and the unused `_resolveUpdatesStream({override})` argument.
- **Public API coupling** — switched to `Stream<Object>?` on `initialize()` (mobile-only; cast to `TaskUpdate` inside the mobile hub module). Happy to follow up with a flutter_gemma-owned update type if you prefer that over `Object`.
- Tests now cover single-sub dual-listener, broadcast identity, hub error fan-out, and clear/reset lifecycle.
- Added dartdoc on the new `initialize()` params + CHANGELOG line under 1.1.0.

Let me know if you'd like any further tweaks before merge.

DenisovAV:
Thanks again @hughesyadaddy — this shipped in **flutter_gemma 1.1.2** 🎉

Your change is live on pub.dev. I integrated it via #345 (re-applied on current `main` with the review fixes — broadcast guard, `onError`/`onDone` hub propagation, the web-layering split, plus a `catchError` on the `_ensureConfigured` chain so a config failure surfaces instead of hanging). You're credited as co-author on the commit and in the CHANGELOG.

Closing this as superseded by #345 (GitHub won't let it merge directly now that the same code is already on `main` — it shows as conflicting). The work itself is in 1.1.2. Really appreciate the contribution and the fast turnaround on the review feedback! 🙏

Issue #323, opened 2026-06-16
Support for Built-in Kotlin


remingtonc:
Opening an issue to track, as this message popped on build.

```
WARNING: Your app uses the following plugins that apply Kotlin Gradle Plugin (KGP): flutter_gemma, flutter_timezone, large_file_handler, shared_preferences_android, url_launcher_android
Future versions of Flutter will fail to build if your app uses plugins that apply KGP.

Please check the changelogs of these plugins and upgrade to a version that supports Built-in Kotlin.
If no such version exists, report the issue to the plugin. If necessary, here is a guide on filing 
an issue against a plugin: https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-app-developers#report-incompatible-kotlin-gradle-plugin-usage-to-plugin-authors
```

Just2sire:
Same issue here: 
WARNING: Your app uses the following plugins that apply Kotlin Gradle Plugin (KGP): flutter_gemma, flutter_gemma_mediapipe
Future versions of Flutter will fail to build if your app uses plugins that apply KGP.

Please check the changelogs of these plugins and upgrade to a version that supports Built-in Kotlin.
If no such version exists, report the issue to the plugin. If necessary, here is a guide on filing 
an issue against a plugin: https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-app-developers#report-incompatible-kotlin-gradle-plugin-usage-to-plugin-authors

If you are a plugin author, please migrate your plugin to Built-in Kotlin using this guide: https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors

DenisovAV:
Quick update: 1.2.2 (`flutter_gemma`) and 1.0.4 (`flutter_gemma_mediapipe`) fix the actual **build break** that the KGP migration introduced on AGP 9 (see #360) — the plugins now compile correctly whether or not `android.builtInKotlin` is enabled.

[...]

PR #343, opened 2026-06-24
fix(android): migrate to Built-in Kotlin (#323)

DenisovAV:
Fixes **#323** — Flutter 3.44+ warns that `flutter_gemma` + `flutter_gemma_mediapipe` apply the Kotlin Gradle Plugin (KGP), and AGP 9 removes support for applying it, so unmigrated plugins **will fail to build on AGP 9+**.

## Change
Both Android modules now follow the [Flutter Built-in Kotlin migration guide](https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors) for plugins that must support **both** AGP 8 and 9:

```groovy
def agpMajor = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize(".")[0] as int
if (agpMajor < 9) {
    apply plugin: "kotlin-android"   // AGP 8.x still needs it
}                                    // AGP 9+: built-in Kotlin takes over
```

- `apply plugin: "kotlin-android"` is now conditional (skipped on AGP 9+).
- The `kotlin { compilerOptions { jvmTarget } }` block moves to the root (was nested in `android {}`).

## Why conditional (not a plain removal)
AGP 8.x has no built-in Kotlin — removing the apply outright would break Kotlin compilation for the majority of consumers still on AGP 8. The conditional form is the guide-recommended approach: applies KGP on AGP 8, skips it on AGP 9. This is what first-party plugins do.

## Verified
- `flutter build apk --release` succeeds on our AGP 8.9 (Kotlin compiles, R8 + packaging pass).
- The KGP warning **persists on AGP 8** (KGP is still applied there — detected at config time, unavoidable while AGP 8 needs it) and **clears on AGP 9+**, where the build would otherwise fail. The real deprecation threat (AGP 9 build failure) is resolved.

## Versions
- `flutter_gemma` 1.1.1 → **1.1.2**
- `flutter_gemma_mediapipe` 1.0.2 → **1.0.3**
- ios/macos podspecs synced (mediapipe podspec also corrected from a stale `0.1.0`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/flutter_gemma/CHANGELOG.md
+++ b/packages/flutter_gemma/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 1.1.2
+- Migrate Android module to Built-in Kotlin — apply KGP only on AGP < 9, ready for AGP 9+ (#323).
+
 ## 1.1.1
 - Declare `libvndksupport.so` for the Android GPU backend — fixes Mali GPU hard-freeze on Android 12+ (#324).
 
--- a/packages/flutter_gemma/android/build.gradle
+++ b/packages/flutter_gemma/android/build.gradle
@@ -74,7 +74,10 @@ android {
 // the `kotlin {}` DSL isn't registered at the Project level for a library
 // subproject — a bare `kotlin {}` block there fails with "Could not find method
 // kotlin()". pluginManager.withPlugin defers the config so it's a no-op when KGP
-// isn't applied, instead of crashing configuration.
+// isn't applied, instead of crashing configuration. On AGP 9 the explicit
+// jvmTarget is not lost: built-in Kotlin derives it from compileOptions
+// (sourceCompatibility/targetCompatibility VERSION_11 above), so the JVM 11
+// target is preserved by a different mechanism.
 pluginManager.withPlugin('kotlin-android') {
     kotlin {
         compilerOptions {
--- a/packages/flutter_gemma/android/build.gradle
+++ b/packages/flutter_gemma/android/build.gradle
@@ -68,11 +68,18 @@ android {
     }
 }
 
-// Built-in Kotlin places compilerOptions in a root `kotlin {}` block (works on
-// both the AGP 8 KGP path and AGP 9 built-in path).
-kotlin {
-    compilerOptions {
-        jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_11
+// Set the Kotlin jvmTarget only once the kotlin-android plugin is actually
+// applied. On AGP < 9 we apply it above, so this runs and configures jvmTarget.
+// On AGP 9 the plugin isn't applied (built-in Kotlin handles compilation) and
+// the `kotlin {}` DSL isn't registered at the Project level for a library
+// subproject — a bare `kotlin {}` block there fails with "Could not find method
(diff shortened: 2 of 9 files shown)


DenisovAV:
Update — verified against AGP 9 and hardened the jvmTarget config.

**Finding:** the migration guide's bare \`kotlin { compilerOptions {} }\` block actually **crashes on AGP 9** for a library subproject like ours — `Could not find method kotlin()` — because the \`kotlin {}\` DSL convenience isn't registered at the Project level when KGP isn't applied (the AGP 9 built-in path). I reproduced this on a real AGP 9.0.1 + Gradle 9.1 build.

**Fix:** the jvmTarget config is now deferred via \`pluginManager.withPlugin('kotlin-android') { kotlin { ... } }\` — it runs only when KGP is actually applied (AGP < 9) and is a no-op otherwise, instead of failing configuration.

**Verified on both paths:**
- **AGP 8.9.1** — release APK builds green; the \`withPlugin\` callback fires and sets \`jvmTarget = JVM_11\` (confirmed via a temporary diagnostic).
- **AGP 9.0.1 + Gradle 9.1, built-in Kotlin on** — both \`:flutter_gemma\` and \`:flutter_gemma_mediapipe\` configure with **no \`kotlin()\` crash**.

**Known limitation:** a full app build on AGP 9 with Flutter 3.44 isn't achievable yet end-to-end (the broader ecosystem — incl. some app-level and third-party plugins — isn't AGP-9-ready; flutter/flutter#181383 is still open). This PR's scope is: stop tripping the KGP deprecation warning on AGP 9, and never crash plugin configuration on either AGP. The full green AGP-9 build will land as the ecosystem catches up.

DenisovAV:
Ran a focused review (Gradle-semantics + release/CLAUDE.md) + full pre-flight. Summary:

**Verified**
- AGP 8.9.1 release APK builds green; the `withPlugin('kotlin-android')` callback fires and sets `jvmTarget = JVM_11`.
- AGP 9.0.1 + Gradle 9.1 (built-in Kotlin on) configures both modules with no `kotlin()` crash.
- `dart analyze` + `flutter test`: core 0 errors / 336 pass; mediapipe 0 errors / 11 pass.
- Versions consistent: `flutter_gemma` 1.1.2 (pubspec + ios/macos podspec + CHANGELOG), `flutter_gemma_mediapipe` 1.0.3 (pubspec + ios podspec + CHANGELOG). Scope = only the two Android plugin modules; the other packages ship native via Native Assets (no Kotlin Gradle module), correctly untouched.

**Addressed in last commit:** clarifying comment that on AGP 9 the explicit `jvmTarget` isn't lost — built-in Kotlin derives JVM 11 from `compileOptions`.

**Known follow-up (non-blocking):** the `buildscript` block still pins `kotlin-gradle-plugin` on the classpath unconditionally. The *apply* is now conditional, but the classpath isn't — it can't be gated by AGP version there (AGP isn't loaded yet in `buildscript`). Fully removing it is a larger buildscript→plugins-DSL modernization that risks the AGP 8 path; tracking separately. Today it's harmless (AGP 9 configures clean), and the #323 goal — no KGP-apply deprecation warning, builds on AGP 8 + 9 — is met.

PR #369, opened 2026-07-10
fix(core): FunctionGemma format — typed args, streaming truncation, chat_template parity (#366, #367)

### LINKED ISSUE #366: FunctionGemma arg parser silently drops array and numeric parameters
### Bug: FunctionGemma arg parser silently drops array and numeric parameters

**Version:** flutter_gemma 1.2.2 · `ModelType.functionGemma` · Android MediaPipe `.task`

**Summary**
`FunctionGemmaCallFormat.parse()` extracts call arguments with:

```dart
// lib/core/parsing/function_gemma_format.dart
final paramRegex = RegExp(r'(\w+):<escape>(.*?)<escape>');
```

That pattern only matches `<escape>`-wrapped **scalar strings**. FunctionGemma
also emits **arrays** and **bare numbers**, and both are silently dropped:

- array:  `extensions:[<escape>pdf<escape>]`  → the `[` after `:` breaks `\w+:<escape>`
- number: `older_than_days:30`, `min_size_mb:100` → no `<escape>` at all

**Reproduce**
A FunctionGemma model returns (verified with a fine-tuned FunctionGemma-270M):

```
<start_function_call>call:move_files{destination:<escape>Documents<escape>,extensions:[<escape>pdf<escape>],min_size_mb:100}<end_function_call>
```

`parse()` yields:

```dart
FunctionCallResponse(name: move_files, args: {destination: Documents})
```

Expected:

```dart
args: {destination: Documents, extensions: [pdf], min_size_mb: 100}
```

`extensions` and `min_size_mb` are gone, so any tool that uses list/number
parameters loses them on-device (in our case `move_files` loses its type/size/
date filters and moves *every* file instead of the matching ones).

**Suggested fix**
Extend the argument parser to handle all three value shapes, e.g.:
- array: `(\w+):\[([^\]]*)\]` then pull `<escape>(.*?)<escape>` items into a `List`
- scalar string: the existing `(\w+):<escape>(.*?)<escape>`
- number: `(\w+):\s*(-?\d+(?:\.\d+)?)` → `num`

parsing arrays/numbers before/independently of the scalar pass so they don't
overlap. Happy to open a PR if useful.

**Workaround** (for others hitting this): read the raw call text from
`chat.fullHistory.last.text` (the plugin stores it via `Message.toolCall`) and
re-parse arguments yourself.

### LINKED ISSUE #367: FunctionGemma tools prompt is rebuilt in Dart and diverges from the model's chat_template property order → degraded output
### Bug: FunctionGemma tools prompt is rebuilt in Dart and diverges from the model's chat template (property order), degrading output

**Version:** flutter_gemma 1.2.2 · `ModelType.functionGemma` · Android MediaPipe `.task`

**Summary**
`_createFunctionGemmaToolsPrompt()` (lib/core/chat.dart) reconstructs the tool
`<start_function_declaration>…<end_function_declaration>` block in Dart, emitting
each tool's `properties` in **insertion/JSON-schema order**:

```dart
properties.forEach((name, schema) { ... });   // schema order
```

But FunctionGemma models are trained/served with a chat template
(`chat_template.jinja`) that renders properties **alphabetically**:

```jinja
{%- for key, value in properties | dictsort -%}   {# sorted #}
```

So the on-device declaration block is off-distribution vs what the model saw in
training, and extraction quality drops.

**Reproduce**
Fine-tuned FunctionGemma-270M, tool `move_files` with schema-order properties
`destination, extensions, name_contains, older_than_days, newer_than_days,
min_size_mb, max_size_mb`.

- Same prompt via HF `tokenizer.apply_chat_template(msgs, tools=…)` (alphabetical
  order) → correct: `move_files{destination:…,extensions:[pdf]}`.
- Same prompt on-device via flutter_gemma (schema order) → the model regresses
  and echoes the request into a string arg:
  `move_files{destination:<escape>Documents<escape>,extensions:[<escape>pdf<escape>],name_contains:<escape>all PDFs to Documents<escape>}`
  (a `name_contains` the HF-template path never produces).

The HF template also differs in other rendered details (it emits
`type:<escape>STRING<escape>`, `items:{…}`, `required:[…]`, `type:<escape>OBJECT<escape>`),
so property order is one concrete divergence, not necessarily the only one.

**Suggested fix**
Render the tools prompt from the **model's own `chat_template.jinja`** (via the
bundled tokenizer / minja) rather than reconstructing it in Dart — the Dart path
will drift from any given FunctionGemma checkpoint's template. As a minimal
interim fix, sort `properties` alphabetically to match `dictsort`.

**Note:** combined with #366 (arg parser dropping arrays/numbers), on-device
FunctionGemma tool calls are effectively unusable for tools with list/number
params or non-alphabetical schemas. Happy to help / open a PR.

DenisovAV:
Two issues from @Ojansen, one root cause: our FunctionGemma **encoder** and **decoder** were written independently, and both drifted from the model's own `chat_template.jinja`. Three rounds of adversarial review and differential fuzzing against that template turned up eight more defects in the same format — all pre-existing, all fixed here.

### #366 — the decoder dropped every non-string argument

`FunctionGemmaCallFormat.parse()` matched one regex, `(\w+):<escape>(.*?)<escape>`. So `call:move_files{destination:<escape>Documents<escape>,extensions:[<escape>pdf<escape>],min_size_mb:100}` parsed to `{destination: Documents}`: `extensions` and `min_size_mb` vanished, and the tool ran on unfiltered input. Never a regression — the regex is string-only in the original FunctionGemma implementation.

### #366, part two — streaming truncated the call at the first `}`

`isFunctionCallComplete()` answered "done" as soon as the buffer ended with `}`, and `chat.dart` parses the first buffer that says so. A `}` closing a nested object, or sitting inside an escaped string, ended the call early:

| Model emitted (well-formed, end tag intact) | Parsed | Leaked to the user as text |
|---|---|---|
| `call:f{a:{x:1},b:2}` | `{a:{x:1}}` — `b` dropped | `,b:2}<end_function_call>` |
| `call:f{msg:<escape>a}b<escape>,x:1}` | `{msg:"a"}` — value cut, `x` dropped | `b<escape>,x:1}…` |

The second row is the common one: **any string argument containing `}`** broke the call. Calls are now delimited by brace depth with `<escape>…<escape>` spans opaque; both regexes are gone and `<end_function_call>` stays optional. The tag itself no longer leaks into chat text or history.

### Eight more, found in review

**Decoder**
- **A bare `None`** — how Jinja renders a null argument — was read as the string `"None"`. A real `"None"` arrives escape-wrapped, so the mapping is collision-free.
- **Ambiguous bodies were silently truncated.** The template never escapes `<escape>` inside a string value, so `{x:"a<escape>b", y:1}` is ambiguous on the wire. We parsed `x:"a"` and dropped `y` — firing the tool with corrupted arguments. The whole call is now refused and the raw text surfaces.

**Encoder**
- **A property with no `type`** silently became `STRING`. The model then quoted numbers back, and a tool expecting `5` received `"5"`. Now the type is inferred from `properties` → OBJECT and `items` → ARRAY (JSON Schema implies it), and a property with neither raises an `ArgumentError` naming its tool.
- **`type: ["string","null"]`** escaped `createToolsPrompt()` as a bare `_TypeError`. Now an `ArgumentError` pointing at `nullable: true`, the template's own structural key.
- **No-argument tools** lost their entire `parameters` block.
- **`items` dropped an empty `required: []`.** The template guards empty `required` at the object level but not inside `items` — an asymmetry in the template itself.
- **`_dictsort` was unstable** past Dart's 32-element threshold, and folded `İ` differently from Python's `str.lower()`.
- **Numbers, booleans and `None` were rendered Dart-style.** They reach the prompt through Jinja, i.e. Python's `str()`: `1e16` became `10000000000000000.0`, `True` became `true`.

### Ground truth

`google/functiongemma-270m-it` is gated, but the identical `chat_template.jinja` ships ungated in `onnx-community/functiongemma-270m-it-ONNX`. Rendering it locally makes the template both the spec and the oracle.

### Where we deliberately diverge

The template sometimes renders what a model cannot use. Five cases are pinned by test rather than copied: a null `description` printed as the literal `None`; an unclosed `parameters:{` with a trailing comma when the schema omits its top-level `type`; a stray sibling key such as `additionalProperties` promoted into a fake parameter; a degenerate property for a non-map schema; and an empty `type:<escape><escape>` where we infer OBJECT or ARRAY instead.

### Verification

- 437 unit tests green (was 402), including a scripted token stream through `generateChatResponseAsync()` — the parser tests call `parse()` with a whole call and never touch the per-token gate, which is exactly how the truncation bug hid behind green tests.
- **Encoder differential:** 584 schemas rendered through the real template and diffed — byte-identical. Seeded fuzz over property counts 0–40, case collisions, structural-key names, nested objects, arrays, unions inside `items`, enums, unicode keys, plus float magnitudes from `1e-9` to `1e100`.
- **Decoder round-trip:** 420 calls rendered by the template's own `format_argument` macro and parsed back. 286 exact; 72 refused (every one carries a literal `<escape>` — see below); 62 integers beyond 2^63 widen to `double`; zero silent mismatches.

### Known format limits

- A string argument containing the literal `<escape>` sentinel cannot be recovered — the template does not escape it. Such calls are refused rather than corrupted.
- Integers beyond 2^63−1 arrive as `double`; Dart has no wider fixed-width int.

### Compatibility

- **Single-property tools are unchanged byte-for-byte.**
- **Multi-property tools now render sorted.** For the base model this is the shape it was trained on. `colabs/functiongemma_finetuning.ipynb` hardcoded its declarations in schema order, so it is updated here too — its three declarations are now byte-identical to what the encoder emits for the example app's tools. **Anyone who fine-tuned with the old notebook must retrain**, or pin `flutter_gemma` below 1.2.3.
- **`args` values are now typed** (`int`/`double`/`bool`/`List`/`Map`/`null`) instead of always `String`. `FunctionCallResponse.args` was already declared `Map<String, dynamic>`, and the gemma4/JSON parsers already returned typed values, so no consumer in this repo breaks.
- **A FunctionGemma tool property must now resolve to a type.** `genkit_flutter_gemma` passes `ToolDefinition.inputSchema` straight into `Tool.parameters`, so a Zod `z.any()`/`z.unknown()` property — which serialises to `{}` — now raises an `ArgumentError` from `ai.generate(...)` instead of being silently treated as a string. Give the property a concrete type. Only `ModelType.functionGemma` is affected; every other model type serialises the schema as JSON and never inspects `type`.

Released as `1.2.3`.

@Ojansen — would you check this against your fine-tuned model? Note the property-order caveat above if you trained with the notebook.

Closes #366
Closes #367


Ojansen:
Tested against the fine-tuned model on-device — **both fixes confirmed green.** 🎯

**Setup.** FunctionGemma-270M fine-tune, Galaxy S22, `.task` via MediaPipe on CPU. To isolate your format change from the monorepo-HEAD build surface, I backported just the two changed files (`core/chat.dart` encoder + `core/parsing/function_gemma_format.dart` decoder) onto published `1.2.2` via a `dependency_override`. Both files dropped in clean — identical imports, every new symbol defined in-file.

**#367 — encoder.** The on-device declaration block your encoder now emits is **byte-identical** to what the model's `chat_template.jinja` renders for our schemas: `dictsort` (case-insensitive) property order, `items:{type:<escape>STRING<escape>}` on every array, `enum` only on the STRING property, `type` last. I also rendered our exact tool schemas through the new `_formatFunctionGemma*` helpers standalone and diffed against the template render — matched, including the array-of-string and enum cases.

And it fixes the actual behavior that made me file #367. Same prompt that used to leak the request text into `name_contains`:

```
prompt: "move all PDFs to Documents"
RAW:    <start_function_call>call:move_files{destination:<escape>Documents<escape>,extensions:[<escape>pdf<escape>]}
```

Clean call, **no `name_contains` leak** — with the training-shaped declaration block the fine-tune behaves exactly like it does under `transformers`.

**#366 — decoder.** The plugin's own `FunctionCallResponse.args` now reads back:

```
move_files {destination: Documents, extensions: [pdf]}
```

The `extensions` array round-trips. On `1.2.2` it was silently dropped (`{destination: Documents}` only).

**End-to-end.** The plan correctly resolved to **117 PDFs → Documents/** with no app-side cleanup — I can now delete the two workarounds we'd been carrying (a raw-text reparser for the dropped args, and a heuristic that stripped the leaked `name_contains`). One root cause, both halves fixed. Thanks for the thorough write-up and the template-as-oracle approach — nice fix. ✅

Ojansen (approved):

Issue #373, opened 2026-07-11
.litertlm FFI: abandoning/cancelling a response stream never cancels the native generation — worker decodes the full turn invisibly; composes with #364's sync teardown into multi-second ANRs


Abdallah01:
**Versions:** found on flutter_gemma 0.16.3; verified still present on current main (`packages/flutter_gemma_litertlm/lib/src/ffi/litert_lm_client.dart`). Android arm64 (Galaxy S25), Gemma-4-E2B `.litertlm`, GPU backend.

## Symptom

When a consumer cancels its `StreamSubscription` on a streaming response (or otherwise abandons the stream — e.g. a newer request supersedes an in-flight one), the native generation is **not** cancelled. The accelerator decodes the entire remaining turn for output nobody will consume. In a high-turnover workload (real-time on-device translation, where scrolling frequently supersedes in-flight batches) this silently consumed roughly **half** of our accelerator time: after working around it app-side, our served-work share went from 23.9 % to 51.2 % on the same workload.

## Root cause (0.16.3 line refs; same structure on current main)

- `_doSendMessageStreamRawOn`'s `StreamController` has **no `onCancel` handler** — a consumer cancel just detaches Dart-side.
- `startVirtualTurn`'s `controller.onCancel` calls `inner?.cancel()` and releases the mutex/turn-flag, but never calls `_cancelOn(conv)`.
- `litert_lm_conversation_cancel_process` is therefore only reachable via an **explicit** `stopGeneration()` — implicit abandonment is silent. Empirically: zero `[LiteRtLmFfi] Generation cancelled` lines across ~1.5 M log lines of our production workload before the workaround, against hundreds of abandoned turns.

## Why it's worse than wasted compute — composition with #364

[...]

PR #375, opened 2026-07-11
fix(litertlm): cancel native decode before tearing a conversation down (#364, #373)

DenisovAV:
Addresses the teardown half of #364 and #373 (both from @Abdallah01, with production ANR evidence). Small, isolate-free, and independently shippable.

## The bug

`litert_lm_conversation_delete` blocks natively in `SessionBasic::~SessionBasic → ThreadPool::WaitUntilDone` — an **unbounded wait** for the in-flight generation to drain. Run on the platform thread under GPU throttle, that tail is tens of seconds → Android input-dispatch ANR + system kill. Reporter's evidence: **14/15 ANR traces** stuck in exactly this stack; one main-thread freeze of **15.5 s**, unblocking the millisecond `Conversation closed` printed. Healthy deletes (idle worker) are 6–8 ms — the cost is entirely the drain.

The fix is to **cancel the native decode before deleting**, so the drain returns in milliseconds.

## What was already correct

Explicit `subscription.cancel()` already reaches native on `main`: the raw stream's `onCancel → _cancelOn` was wired in the 0.16.2 virtual-session refactor, and Dart propagates `cancel()` through the `async*`/`.map()` chain. This PR closes the two teardown paths that still drained **without** cancelling first.

## Changes

1. **`LiteRtLmConversationHandle.close()`** — cancel the live decode before `_deleteConversation`. `session.close()` mid-generation on the single-session path hit the full drain; the virtual path already guarded this. `_cancelOn` is lock-free, so it interrupts a generation holding `_nativeMutex`.
2. **`startVirtualTurn.onCancel`** — cancel the native turn **explicitly**, not only via `inner`. A consumer cancel that lands while `onListen` is still awaiting the mutex or the create left `inner` null → the cancel never reached native, and `onListen` would then start a generation nobody consumes. A flag now makes `onListen` bail before launching that orphan turn.
3. **Docs** — `getResponseAsync`/`stopGeneration`: abandoning a stream (dropping the subscription without cancelling) cannot be intercepted; the accelerator decodes the whole turn. Consumers must `cancel()` or `stopGeneration()`. This matches @Abdallah01's app-side workaround.

## Invariants preserved (from #365/#372)

Cancel stays **lock-free** (must interrupt a mutex-holding generation, never await the mutex). Shutdown ordering unchanged — cancel only shortens the delete drain. No new UAF: on the virtual path `_cancelOn` targets `_virtualConv`, deleted only under the mutex being released. Double-cancel (`cancelVirtualTurn` + `inner.cancel()`) is idempotent natively.

## Verification

- Unit: 441 core + 31 litertlm tests green; analyze clean. (This teardown-ordering + FFI code isn't unit-testable without mockable bindings, same as #365/#372 — verified by structure + the race analysis above.)
- **On-device needed** — @Abdallah01 offered Android arm64 / Galaxy S25 / Gemma-4-E2B `.litertlm` / GPU. The check: after a mid-decode cancel, `[LiteRtLmFfi] Generation cancelled` must appear (reporter saw **zero** across 1.5M production log lines) and `Conversation closed` must follow within ms, not seconds.

## Not in scope

The **worker-isolate** move for the remaining synchronous-FFI UI jank (the larger half of #364) is a separate, phased change. This PR is the teardown/ANR half, which the reporter notes is what collapses the multi-second stall.

@Abdallah01 — could you run this branch on your workload and confirm the ANR is gone and `Generation cancelled` now appears?

### DIFF
--- a/packages/flutter_gemma/CHANGELOG.md
+++ b/packages/flutter_gemma/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 1.2.3
-- Cancel native `.litertlm` decoding before tearing a session down — avoids a multi-second platform-thread stall / ANR while `conversation_delete` drains an abandoned generation (#364, #373).
+- Cancel native `.litertlm` decoding before session teardown to avoid an ANR (#364, #373).
 - Fix FunctionGemma parser dropping array, number, boolean and object arguments (#366).
 - Fix FunctionGemma streaming truncating a tool call at the first `}` (#366).
 - Fix FunctionGemma tools prompt to match the model's `chat_template` (#367).
--- a/packages/flutter_gemma/CHANGELOG.md
+++ b/packages/flutter_gemma/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## 1.2.3
+- Cancel native `.litertlm` decoding before tearing a session down — avoids a multi-second platform-thread stall / ANR while `conversation_delete` drains an abandoned generation (#364, #373).
 - Fix FunctionGemma parser dropping array, number, boolean and object arguments (#366).
 - Fix FunctionGemma streaming truncating a tool call at the first `}` (#366).
 - Fix FunctionGemma tools prompt to match the model's `chat_template` (#367).
--- a/packages/flutter_gemma/lib/flutter_gemma_interface.dart
+++ b/packages/flutter_gemma/lib/flutter_gemma_interface.dart
@@ -442,12 +442,16 @@ abstract class InferenceModelSession {
 
   /// Streams generated tokens.
   ///
-  /// To stop generating early, either `cancel()` the [StreamSubscription] or
-  /// call [stopGeneration]. On the `.litertlm` FFI path both reach the native
-  /// engine and stop it decoding. Silently *abandoning* the stream (dropping
-  /// the subscription without cancelling, so it is only GC'd) cannot be
-  /// intercepted — the accelerator keeps decoding the whole turn for output no
-  /// one consumes, wasting compute and delaying the next turn's teardown.
+  /// To stop generating early, prefer [stopGeneration] — it reaches the native
+  /// engine and stops it decoding on every engine. `cancel()`ing the
+  /// [StreamSubscription] also stops native decoding on the `.litertlm` FFI
+  /// engine, but on other engines it only detaches the Dart side while native
+  /// keeps decoding, so [stopGeneration] is the portable choice.
+  ///
+  /// Silently *abandoning* the stream (dropping the subscription without
+  /// cancelling, so it is only GC'd) cannot be intercepted on any engine — the
+  /// accelerator keeps decoding the whole turn for output no one consumes,
+  /// wasting compute and delaying the next turn's teardown.
   Stream<String> getResponseAsync();
 
   Future<int> sizeInTokens(String text);
--- a/packages/flutter_gemma/lib/flutter_gemma_interface.dart
+++ b/packages/flutter_gemma/lib/flutter_gemma_interface.dart
@@ -440,12 +440,23 @@ class SessionMetrics {
 abstract class InferenceModelSession {
   Future<String> getResponse();
 
+  /// Streams generated tokens.
+  ///
+  /// To stop generating early, either `cancel()` the [StreamSubscription] or
+  /// call [stopGeneration]. On the `.litertlm` FFI path both reach the native
+  /// engine and stop it decoding. Silently *abandoning* the stream (dropping
+  /// the subscription without cancelling, so it is only GC'd) cannot be
+  /// intercepted — the accelerator keeps decoding the whole turn for output no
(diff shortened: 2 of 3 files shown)


Abdallah01:
Ran this on our production workload as requested — **the ANR is gone and `Generation cancelled` now appears exactly as intended.** Details below.

**Test build (disclosure):** our toolchain is Dart 3.11.5 and the branch tip requires language version 3.12 (the kernel compiler rejects it even with dependency overrides), so we tested **hosted 0.16.3 with this PR's `litert_lm_client.dart` hunks backported verbatim** — `_cancelOn` before `_deleteConversation` in `close()`, the `cancelledByConsumer` onListen bail, and `cancelVirtualTurn` in `onCancel`. Galaxy S25, GPU backend, Gemma "4-E2B"-class model, Flutter profile build (md5-verified install).

**Workload:** ~10 minutes of continuous live translation — sustained streamed generations back-to-back — driven into the warm regime (skin ~41 °C, sustained GPU load; the exact conditions where the old teardown drain stretched to ANR length), with deliberate mid-generation teardowns:

- app backgrounding ×2
- full session dispose + recreate ×1
- conversation close + recreate ×2
- frequent consumer-side stream cancellations throughout (in-flight generations superseded by newer requests)

**Results:**

- `[LiteRtLmFfi] Generation cancelled` — **39 occurrences**. For calibration: this marker was **zero across ~1.5M log lines** of the same workload on unpatched 0.16.3.
- Every cancel paired with a `Conversation closed`: **min/median 0 ms, p95 68 ms, max 73 ms** cancel→close latency. On unpatched 0.16.3 a close with generation in flight blocked in `litert_lm_conversation_delete` → `~SessionBasic` → `ThreadPool::WaitUntilDone` for seconds to tens of seconds under GPU throttle.
- **ANR: zero.** No `Input dispatching timed out`, no ANR dialogs, no abnormal process exits (`logcat` + `dumpsys activity exit-info` both clean). UI stayed responsive through every teardown, including taps landed mid-teardown.
- To keep the A/B clean we **disabled our app-side workaround** (explicit `stopGeneration()` before teardown) for this run; its log marker never fired, so all 39 cancellations are attributable to this PR's cancel paths alone.

Thanks for the fast turnaround on this. This closes out the worst UX cliff we had on-device.

DenisovAV:
Thanks for the clean A/B — disabling your app-side `stopGeneration()` workaround so all 39 cancellations are attributable to the PR is exactly the isolation that makes this conclusive. Much appreciated.

One clarification so your result carries over cleanly: the fix shipped in `flutter_gemma_litertlm` **1.0.3** is byte-identical to the `litert_lm_client.dart` hunks you backported (`_cancelOn` before `_deleteConversation` in `close()`, the `cancelledByConsumer` onListen bail, and `cancelVirtualTurn` in `onCancel`), so your verification stands for 1.0.3 as-is.

On the toolchain wall you hit: the 1.x line has a hard `sdk: '>=3.12.0'` / `flutter: '>=3.44.0'` floor. That's a **language-version** gate enforced by the kernel compiler, which is why `dependency_overrides` can't relax it — overrides only change which package *version* resolves, not the SDK/language floor. The floor isn't arbitrary: it was raised at the 1.0 cut for `large_file_handler 0.5.0` + a dart2wasm-clean web graph, so we can't trivially lower it without regressing those.

Could you tell us **what's keeping you on Dart 3.11.5** — a specific dependency that isn't 3.12-compatible yet, a pinned CI/toolchain image, or something else? If it's a blocking dependency we may be able to help unblock the Flutter 3.44 move; if a chunk of users are genuinely stuck below 3.12, that changes whether a 0.16.x backport of these ANR fixes is worth cutting. Your answer decides which path we take.

Abdallah01:
Glad the isolation was useful — one small correction for the record: we didn't disable the app-side `stopGeneration()` workaround for the A/B; it stayed in place and its failure-path marker simply never fired, which is what made all 39 cancellations attributable to the PR's paths.

Good to know 1.0.3 is byte-identical to the backported hunks — and we can also confirm directly: we've since moved to Flutter 3.44.6 / Dart 3.12 and re-ran the same teardown protocol on released 1.2.3 + flutter_gemma_litertlm 1.0.3 (real device, ~10 min hot-regime, backgrounding / tab close / navigation / scroll-supersede): cancellations fire only on genuine aborts as expected post-`mutexHeld` guard, 0 ANRs, clean exit-info.

To answer the question directly: nothing structural was keeping us on 3.11.5 — it was sequencing (validate the fix on the toolchain we were on, then take the SDK bump as its own change). We're on 3.12 now, so **we don't need a 0.16.x backport**; please don't weigh our earlier constraint in that decision.

One heads-up if you do cut one for other users: #379 (the cancel-after-delete UAF in `Conversation::CancelProcess`) rides with the onCancel path, so a 0.16.x backport of these fixes would presumably carry the same hazard — worth including whatever guard lands for #379 in the same cut.

Issue #379, opened 2026-07-12
SIGSEGV (use-after-free) in Conversation::CancelProcess when model is closed while a session stream is attached (1.2.3 / litertlm 1.0.3)


Abdallah01:
## Summary

Native SIGSEGV (use-after-free) in `litert::lm::Conversation::CancelProcess()` on **flutter_gemma 1.2.3 / flutter_gemma_litertlm 1.0.3** (Android arm64, Galaxy S25, profile build, Flutter 3.44.6) when an `InferenceModel` is closed/re-created while an `InferenceModelSession`'s `getResponseAsync()` stream is still attached. Follow-up to the cancel-before-teardown work in PR 375 (#373/#364) — the new onCancel→native-cancel path is what makes this reachable.

## Crash

```
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x5d627574535b0065 (read)
backtrace:
  #00 pc 009d5544  base.apk!libLiteRtLm.so (litert::lm::Conversation::CancelProcess()+4) (BuildId: 2b827d146c50bb4d6a6212805a2cf06b)
  #01-#15 base.apk (Dart AOT frames)
  #16+    libflutter.so … android MessageQueue (platform thread)
```

`CancelProcess()+4` faulting on a read of `this`-adjacent memory, and the fault address is ASCII string bytes — the conversation object was freed and its heap block already reused for string data when the cancel dereferenced it. `ApplicationExitInfo`: `reason=5 (APP CRASH(NATIVE)) status=11`.

## Sequence (observed, ~15 min into a session)

1. Long generation in flight on the live session (`getResponseAsync` streaming).
2. Android critical memory pressure while backgrounded → app decides to unload weights.
3. On resume, app re-warms: it closed the old `InferenceModel` (at the time **without** closing the `InferenceModelSession` first) and created a fresh model.
4. SIGSEGV on the platform thread: a cancel for the orphaned session's stream ran `litert_lm_conversation_cancel_process` on the conversation that the model-level teardown had already deleted.

## Reading of the race (from package source)

- The **session-level** close path handles an in-flight turn safely — `litert_lm_client.dart` defers the conversation teardown to the turn's cleanup ("deleting the pointer now would be a use-after-free") and the virtual-turn `onCancel` is guarded by `mutexHeld` + the `_virtualActiveToken` check.
[...]

PR #380, opened 2026-07-12
fix(litertlm): use-after-free in Conversation::CancelProcess on model-close-mid-stream (#379)

DenisovAV:
Fixes a native SIGSEGV (use-after-free) reported by @Abdallah01 in #379 on shipped `flutter_gemma_litertlm 1.0.3`.

### The bug
When an `InferenceModel` is closed/re-created while a session's `getResponseAsync()` stream is still attached, the conversation is deleted, but a later onCancel for that stream still runs `_cancelOn` on the freed pointer:
- raw-stream hook `controller.onCancel = () => _cancelOn(conv)` captures the pointer by value with no liveness check;
- the virtual-turn path's `_virtualConv` / `_virtualActiveToken` still describe the freed conversation, so `cancelVirtualTurn` passes its guards and calls `_cancelOn` on a dangling pointer.

`_cancelOn` → native `litert_lm_conversation_cancel_process` dereferences freed memory → `SIGSEGV` in `litert::lm::Conversation::CancelProcess`. The #375 cancel-before-teardown path is what makes it reachable (Android arm64, Galaxy S25, GPU, ~15 min into a session under memory pressure).

### The fix
A live-conversation registry (`_liveConvs`): `_createRawConversation` adds the pointer, `_deleteConversation` removes it, and `_cancelOn` no-ops on any pointer not in the set — so a cancel that races or follows teardown can never dereference a freed conversation. This closes both the raw-stream and virtual-turn cancel paths with one guard.

### Verification
- New unit test `test/ffi/cancel_after_delete_test.dart` (registry drops a deleted conversation → late cancel no-ops); full litertlm suite green (32).
- macOS on-device smoke (`litertlm_ffi_test.dart`) with the guard applied: 24/24 — no regression to normal cancel/close/streaming.
- App-side workaround (unchanged behavior for correct callers): `await session.close()` before closing/re-creating the model.

The native Android crash repro is device/heap-timing specific; @Abdallah01 offered to validate a fix build on the exact setup from the #375 run.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -137,7 +137,7 @@ Core has NO pigeon (dropped at the 1.0 cut; its value types are hand-written in 
 - **MediaPipe Web**: v0.10.27, Android/iOS: v0.10.33
 - **LiteRT-LM**: native libs from `native-v0.13.1-a` GitHub Release. Android tarball bundles the Qualcomm QNN dispatch stack and Windows tarball bundles Intel NPU dispatch (`LiteRtDispatch.dll` + OpenVino runtime + TBB) for `PreferredBackend.npu` (Qualcomm Snapdragon / Intel LunarLake/PantherLake). MTP (speculative decoding) support for Gemma 4 (#318 MTP crash fixed). (native-v0.13.1-a restores the NPU dispatch libs accidentally omitted from native-v0.13.1 — #155.)
 - **large_file_handler**: `^0.5.0` (core dep; 0.5.0 declares all 6 platforms — needed for pana platform support + the dart2wasm-clean web graph)
-- **Current Version**: core `flutter_gemma` `1.2.3`, `flutter_gemma_rag_sqlite` `1.1.0`, `flutter_gemma_rag_qdrant` `1.1.0`; `flutter_gemma_litertlm` `1.0.3`, `flutter_gemma_mediapipe` `1.0.4`, `flutter_gemma_embeddings` `1.0.1`; `flutter_gemma_agent` `0.1.0` (new)
+- **Current Version**: core `flutter_gemma` `1.2.3`, `flutter_gemma_rag_sqlite` `1.1.0`, `flutter_gemma_rag_qdrant` `1.1.0`; `flutter_gemma_litertlm` `1.0.4`, `flutter_gemma_mediapipe` `1.0.4`, `flutter_gemma_embeddings` `1.0.1`; `flutter_gemma_agent` `0.1.0` (new)
 - **0.15.2**: embedding unified on LiteRT C API via Dart FFI on all native platforms (Android + iOS + Desktop). Drops `localagents-rag` JVM dep on Android and the separate TFLite C 0.12.7 tarball on Desktop; `TensorFlowLiteC` pod no longer needed on iOS. Single source of truth for `TaskType.prefix` in Dart, fixes cross-platform embedding drift (#264).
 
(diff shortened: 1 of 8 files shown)


Abdallah01:
Validated this branch on the exact setup from the #375/#379 run — Galaxy S25 (Android arm64, SM8750), GPU backend, `flutter_gemma 1.2.3` + `flutter_gemma_litertlm` from `fix/379-cancel-after-delete-uaf` — and, importantly, with the app-side workaround (`await session.close()` before model close/re-create) **removed**, so the vulnerable path from the report — model close/re-create while a session's `getResponseAsync()` stream is still attached — is exactly what runs, and the `_liveConvs` guard stands alone.

**Protocol** (the same chain that produced the original SIGSEGV, forced deliberately): drive sustained generation → `am send-trim-memory <pkg> RUNNING_CRITICAL` mid-generation → model weights evicted → model re-created on demand with the prior session's stream still attached → continued streaming/cancel traffic against the now-deleted conversation. Two independent cycles (separate process launches).

**Results: 2/2 clean.**
- 0 SIGSEGV / 0 native crash blocks; process pid stable through both cycles (no silent restart).
- 45 native cancel operations executed across the two runs (engine `CancelProcess` log lines) — none dereferenced a freed conversation.
- Generation resumed at normal rates after each re-create; no regression observed in normal streaming/cancel/close behavior.
- For reference: this same chain on shipped `1.0.3` produced the `litert::lm::Conversation::CancelProcess` SIGSEGV within one cycle.

**Desk-review notes on the guard:** ordering is right (`_deleteConversation` removes from `_liveConvs` *before* the native delete; `_cancelOn`'s check-then-call is synchronous with no await window), and `shutdown()` cleans the registry for both orphaned handles (via `h.close()`) and `_virtualConv` — so the raw-stream and virtual-turn late-cancel paths are both covered.

One small adjacent observation, take or leave: in `shutdown()`, `_virtualConv` is passed to `_deleteConversation` **without a preceding `_cancelOn`** — unlike `LiteRtLmConversationHandle.close()`, which cancels first precisely because `conversation_delete` blocks in `~SessionBasic → ThreadPool::WaitUntilDone` until the generation drains (the #364 mechanism, per that method's own comment). If a virtual turn is in-flight when `shutdown()` runs, the delete inherits that drain. A symmetric `_cancelOn(vc)` before the delete would close it. Not a blocker for this fix.

DenisovAV:
Thanks for the forced-repro validation — driving `send-trim-memory RUNNING_CRITICAL` mid-generation to evict+re-create with the stream attached, workaround removed, 2/2 clean with 45 native cancels and none dereferencing a freed conversation, vs SIGSEGV-in-one-cycle on 1.0.3: that's exactly the before/after that closes this out. Much appreciated.

And good catch on the desk review. You're right that `shutdown()` deleted `_virtualConv` without a preceding `_cancelOn` — asymmetric with `handle.close()`, so an in-flight virtual turn made `conversation_delete` inherit the `~SessionBasic → ThreadPool::WaitUntilDone` drain (the #364 ANR). I folded a symmetric `_cancelOn(vc)` before the delete into this PR; it's safe post-registry since `vc` is still in `_liveConvs` at that point, so the cancel reaches native and the delete returns fast. It's on the shutdown path — orthogonal to the model-close/re-create chain you stress-tested — so no need to re-run, but you're welcome to if convenient. Shipping in `flutter_gemma_litertlm` 1.0.4.

PR #407, opened 2026-08-03
feat(agent): image input in an agent turn + MCP 2025-11-25

DenisovAV:
## What

- `AgentLoop.run(chat, userMessage, {Uint8List? imageBytes})` and `AgentSession.ask(userMessage, {imageBytes})` now accept an optional image. When present, the turn is built with `Message.withImage` instead of a plain text `Message`, so the model sees the photo alongside the text and can dispatch skills based on what it saw.
- `McpClient` advertises MCP protocol revision `2025-11-25` (was `2025-06-18`), pinned by a test that asserts what goes out in the `initialize` request.

## Requirements

The image only reaches the model when the chat is multimodal — build the session with `supportImage: true` (see `AgentSession.fromModel`) and back it with a vision-capable model. Without that the image is ignored downstream; this is documented on both new parameters.

## Notes

- The fake MCP server in tests still replies with `2025-06-18`. That is deliberate: the client advertises a revision and the server may negotiate down, so the test asserts the outgoing value, not the reply.
- `flutter_gemma_agent` bumped 0.1.0 → 0.2.0.

## Verification

- `flutter test` in `packages/flutter_gemma_agent`: 198 passed.
- `flutter analyze`: no issues.


DenisovAV:
Superseded by #427. This PR was based on agent 0.2.0, before the 0.2.2 loop consolidation (#417) rewrote AgentLoop.run into a StreamController driver over core's generateChatResponseWithTools — so it no longer merged. #427 rebuilds both features (image-in-turn + MCP 2025-11-25) on current main with the same tests, credited via Co-authored-by.

PR #409, opened 2026-08-04
fix(litertlm): drop thinking output from the KV cache on FFI

think2execute:
## Problem

`litert_lm_conversation_config_set_filter_channel_content_from_kv_cache` is declared in `native/litert_lm/include/engine.h` and bound in `litert_lm_bindings.dart`, but nothing on the FFI path ever calls it.

The web path already pairs the two — `litert_lm_web_inference.dart`:

```dart
filterChannelContentFromKvCache: enableThinking ? true : null,
```

`_createRawConversation` in `litert_lm_client.dart` sets `session_config`, `system_message`, `tools`, `enable_constrained_decoding` and `messages`, but not the channel filter, and it had no `enableThinking` parameter to key it off. So on the five native platforms thinking output stays in the KV cache and is re-prefilled on every later turn of the same conversation, while web does not have that cost.

`enableThinking` is already threaded as far as `_chatOn` / `_chatRawOn`, where it becomes per-message `extra_context` — it just never reached the conversation config.

## Change

Threads `enableThinking` from `createSession` down through `createConversationHandle` and `startVirtualTurn` into `_createRawConversation`, and sets the filter when thinking is on.

Defaults to `false` throughout, so nothing changes for sessions that do not enable thinking.

## Why it matters

Measured on a Galaxy M33 (Exynos 1280, Android 15) with Gemma 4 E2B and tools enabled, over a scripted 10-turn conversation run 3×:

- context grows ~109 tokens/turn, and thinking is retained in the cache across all of it
- a warm turn's time-to-first-token is ~4.4s against a ~14.4s cold turn 1, so prefill is already the cheap part — but it grows with every turn that keeps thought in the cache

Turning thinking off entirely cut a 10-turn conversation from 283s to 211s, which is the same cost from the other direction: reasoning that is paid for repeatedly and never read.

## Notes

- No behaviour change for `enableThinking: false`
- `dart analyze` on the two changed files reports 235 issues both before and after this change (all pre-existing `override_on_non_overriding_member` warnings)
- Version bumped to 1.3.2 with a one-line CHANGELOG entry per the repo convention

## Follow-up, not in this PR

`enableThinking` is a `final` field on the session, so it cannot vary per message even though the native layer accepts `extra_context` per message. For a tool-calling chat turn that means the round which only emits a tool-call envelope pays full chain-of-thought for it. Happy to open a separate issue if that is worth pursuing.

### DIFF
--- a/packages/flutter_gemma_litertlm/CHANGELOG.md
+++ b/packages/flutter_gemma_litertlm/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 1.3.2
+- Drop thinking output from the KV cache on FFI, as web already does.
+
 ## 1.3.1
 - Clearer engine-create error for GPU-only `.litertlm` models run on CPU (#390).
 
--- a/packages/flutter_gemma_litertlm/lib/src/ffi/ffi_inference_model.dart
+++ b/packages/flutter_gemma_litertlm/lib/src/ffi/ffi_inference_model.dart
@@ -130,6 +130,7 @@ class FfiInferenceModel extends InferenceModel with CloseNotifier {
         topP: topP,
         seed: randomSeed,
         maxOutputTokens: maxOutputTokens,
+        enableThinking: enableThinking,
       );
       gemmaLog(
         '[FfiInferenceModel/perf] createConversation (FFI): ${sessionSw.elapsedMilliseconds - beforeConv}ms',
@@ -658,6 +659,7 @@ class _VirtualConversationHandle implements ConversationHandle {
         seed: seed,
         extraContext: extraContext,
         maxOutputTokens: maxOutputTokens,
+        enableThinking: enableThinking,
       )) {
         final chunkText = LiteRtLmFfiClient.extractTextFromResponse(rawChunk);
         assistantText.write(chunkText);
--- a/packages/flutter_gemma_litertlm/lib/src/ffi/litert_lm_client.dart
+++ b/packages/flutter_gemma_litertlm/lib/src/ffi/litert_lm_client.dart
@@ -873,6 +873,7 @@ class LiteRtLmFfiClient {
     double? topP,
     int seed = 1,
     int? maxOutputTokens,
+    bool enableThinking = false,
   }) {
     // The handle must be registered before the guard lifts: otherwise shutdown()
     // can clear _handles between the create and the add, leaving a live
@@ -896,6 +897,7 @@ class LiteRtLmFfiClient {
           topP: topP,
           seed: seed,
           maxOutputTokens: maxOutputTokens,
+          enableThinking: enableThinking,
         ),
       );
       gemmaLog('[LiteRtLmFfi] Conversation created');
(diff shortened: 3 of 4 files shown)


think2execute:
Correcting the "why it matters" section above: I have since tried to measure this change on device and **cannot yet show that it helps**. The parity argument stands, but I do not want the numbers in the description read as evidence for the fix.

Two problems with my measurement, both mine:

1. I compared a baseline captured on a cool device against a run on hardware that had been under continuous load for hours. Thermal drift alone covers the difference I saw (median TTFT over later turns moved 2924ms → 2640ms, about −10%, against a run-to-run spread of the same order).

2. My harness re-runs the same conversation three times in one process, and I found afterwards that the vector store failed to open at startup, so retrieval was dead for those runs. Any median I took across them is unreliable.

Separately, the dominant cost in my own app turned out to be something else entirely — a prompt instruction of mine that made the model re-issue an identical tool call within a single turn, costing a full extra generation. That is fixed on my side and had nothing to do with this package.

I will re-measure properly (interleaved A/B, same thermal state, working retrieval) and post the result either way. If it comes back flat I will say so here rather than leave the impression it was validated.

The change itself is unchanged and still correct as parity with the web path: `filterChannelContentFromKvCache` is set there whenever thinking is on, and the FFI path had no way to set it at all.

think2execute:
Measured it properly now, and the result argues **against** enabling this implicitly. Posting in full because it contradicts what I wrote when I opened the PR.

Setup: Galaxy M33 (Exynos 1280, Android 15), Gemma 4 E2B, tools enabled, thinking on. Same device, same build, only this package swapped between runs.

**A fixed 10-turn script, where each turn is largely self-contained** — 2 runs each, reproducible to ~3%:

| | without filter | with filter |
|---|---|---|
| 10-turn total | 342.0s | **212.1s (-38%)** |

That looks like a large win. It is not the whole picture.

**A genuinely conversational suite, where turns depend on earlier ones** — 10 cases, live multi-turn:

| | without filter | with filter |
|---|---|---|
| pass rate | 8/10 | **6/10** |
| follow-up referencing two turns back | PASS (52s) | **FAIL** |
| one case | 18.7s | **wedged, 360s timeout** |
| two later cases | ran | skipped, suite budget exhausted |

The failing case asks a question whose subject was established two turns earlier. With the filter on, the model answered "Could you please tell me what you are you are referring to?" — it had lost the referent. Several other cases got 3-8x slower, which reads like the model floundering without its earlier reasoning rather than any prefill cost.

So the two measurements disagree, and the disagreement is the finding: dropping thought from the KV cache is cheap when nothing needs it and expensive when something does. My first probe only exercised the cheap case.

**Suggestion for the PR:** rather than tying the filter to `enableThinking`, expose it as its own flag so a consumer can choose. Web currently sets `filterChannelContentFromKvCache: enableThinking ? true : null`, so the same trade-off may be silently affecting web today — that is worth checking independently of this change.

Happy to rework the PR that way, close it, or leave it as-is for you to judge. I would not merge it as written.

PR #450, opened 2026-08-20
fix(core): stop claiming background_downloader's updates stream from the host

### LINKED ISSUE #445: Do not register global listener in the `background_downloader`
When `flutter_gemma` registering an internal listener on the `FileDownloader.updates` stream from the `background_downloader` it is effectively breaking any other code or packages that want to get updates about their downloads.

Some context on this issue:
https://github.com/781flyingdutchman/background_downloader/issues/707

Instead of a global listener `flutter_gemma` perhaps should register a callback for a specific downloading "group".

DenisovAV:
Closes #445. Reported by @ekuleshov, diagnosed by @781flyingdutchman — the author of `background_downloader`.

## The failure

`FileDownloader().updates` is a **single-subscription** controller — `var updates = StreamController<TaskUpdate>()` in `base_downloader.dart`. `SmartDownloader` listened to it through `asBroadcastStream()`, which still takes the one subscription the process has.

So any later `FileDownloader().updates.listen(...)` in the host app, or in any other package, threw:

```
Bad state: Stream has already been listened to.
```

Nothing in flutter_gemma failed. The **host** did — merely depending on this package made `background_downloader` unusable for the app's own downloads.

## The fix

Register callbacks scoped to our own task group instead. A strict narrowing, not a workaround: `_emitStatusUpdate` consults `groupStatusCallbacks[task.group]` **before** `updates.hasListener`, so we receive exactly what we received before, minus other people's tasks — which every listen site here already discarded by `taskId`.

The registration is re-established on **every** resolve, not memoized. `FileDownloader().destroy()` clears the group callback map and never touches our controller, so "my controller is open" is a different fact from "I am still registered" — using one as the other left a fan-out that looked healthy and delivered nothing, permanently, with no error.

`clearConfiguration()` closes the fan-out and raises a typed `DownloadUpdatesReleasedException` that is **terminal**, not retried — a dispose used to become up to `maxRetries` restarts of a multi-gigabyte download, each re-registering our callbacks two seconds later, and in every gap our tasks spilled into the host's stream, which is #445 in reverse. The release leaves a silent sink so an in-flight native task can never reach the host.

## Priority: 10 was wrong, but 0 is worse on Android

`priority: 10` with a documented range of `0..10, 0 highest` and a package default of 5. Two different stories per platform:

**iOS** maps it straight onto URLSession — `urlSessionDownloadTask.priority = 1 - task.priority / 10` — so 10 produced `0.0`, below `URLSessionTask.lowPriority`, for a multi-gigabyte download the user is watching. Now `0`.

**Android** uses priority for exactly two things: the holding queue (never enabled here) and `expedited = priority < 5`. Expedited is the wrong trade for a 2–4 GB transfer, twice over:

- it **shortens** the OS execution guarantee — `JobSchedulerService` gives a regular job `RUNTIME_MIN_GUARANTEE_MS` = 10 minutes and an expedited one `RUNTIME_MIN_EJ_GUARANTEE_MS` = 3, and caps expedited at the regular figure. AOSP's own comment there says expedited jobs "shouldn't be used for long pieces of work";
- and it currently hangs the download outright. WorkManager caps a task at 9 minutes; `background_downloader` survives that by pausing and **re-enqueuing with a 1-second delay**. The re-enqueue sets `setInitialDelay` and, below priority 5, `setExpedited` on the same builder — and `WorkRequest.Builder.build()` throws `IllegalArgumentException: Expedited jobs cannot be delayed`. The throw is swallowed to a `Log.w`, `doEnqueue`'s `false` is discarded, and the runner returns `TaskStatus.paused` anyway.

So Android gets `5`, the highest non-expedited value. Against the old `10` that changes nothing observable on Android, which is the point: `10` cost nothing and anything below `5` would.

Keyed off `Platform.isAndroid`, not `defaultTargetPlatform` — a host can legitimately override the latter to preview Cupertino, and doing that in a debug build on a real Android phone would hand this the iOS answer.

**This is not news to the package author.** He has three PRs open on `background_downloader`, all from the day he commented here: [#708](https://github.com/781flyingdutchman/background_downloader/pull/708) documents the single-subscription stream, [#709](https://github.com/781flyingdutchman/background_downloader/pull/709) fixes the expedited-delay crash, and [#710](https://github.com/781flyingdutchman/background_downloader/pull/710) adds the missing `setUserInitiated(true)` — his suggestion of `priority: 0` was premised on UIDT, which the shipped code never actually requested. When those land, the priority choice here is worth revisiting; the dartdoc says so.

## Also fixed, all found by review of this PR

- `TaskStatus.paused` and `enqueued` re-arm the resume watchdog. The 9-minute slice emits `paused → enqueued → running`, and a blanket `_cancelResumeWatchdog` ran before the switch — so a deferred re-enqueue left the download hanging unbounded with no error.
- `TaskStatus.notFound` on the reattach path reports `DownloadError.notFound()`. It is a FINAL state; re-arming a watchdog for it turned a plain 404 into a 90-second stall reported as a network timeout.
- The reattach watchdog does not arm on an already-settled completer. `taskId` is a deterministic hash of (base, directory, filename), so a stale timer would cancel a **new** download of the same model.
- Progress no longer snaps to 0%. `background_downloader` signals state through negative values on the progress channel (`-1` failed … `-5` paused) and both listeners clamped them.
- The download notification is scoped with `configureNotificationForGroup`. The bare `configureNotification` installs a `taskOrGroup: null` **default** that every task in the process falls back to, including the host's — the same overreach as #445.
- The default `foreground: null` path no longer writes `runInForegroundIfFileLargerThan`, a process-wide setting that survives reboot and is shared with the host. It could never take effect for us there (that branch configures no notification, and the platform needs one before starting a foreground service), so it bought nothing and silently overwrote the host's choice. The docs promising ">500MB automatically uses a foreground service" are corrected — that has not been true since #357.
- The example's own resume test listened on `FileDownloader().updates` for one of our tasks and would now receive nothing.
- Core podspecs were four and five releases behind `pubspec.yaml` (1.5.5 and 1.4.0); mediapipe's was at 1.0.4 against 1.0.5.

## Tests

The first suite could not fail on the mechanism: deleting `registerCallbacks` outright left all six green. Delivery now goes through `FileDownloader().downloaderForTesting`, a `@visibleForTesting` getter on the exported class, so an update travels through the real dispatcher.

The enqueue end was unpinned too — changing the task's `group`, or reverting `priority` to 10, or dropping progress updates, left all 636 tests passing. `buildModelDownloadTask` is extracted alongside the file's five other pure helpers, and its `group` is asserted against the **same constant** `registerCallbacks` uses, which is what makes it a drift test rather than a tautology.

Verified by mutation: group drift, priority 10, status-only updates, a hub that does not release the fan-out, and unregistering instead of absorbing on release each fail exactly one test. Removing `registerCallbacks` fails four; restoring the memoized registration fails one.


packages/flutter_gemma/lib/mobile/smart_downloader.dart
```diff
@@ -142,15 +234,90 @@ ResumeAction decideFailedDownloadAction({
...
+
+  /// Scheduling priority for model downloads — and it must differ per platform.
+  ///
+  /// background_downloader documents `0 <= priority <= 10 with 0 being the
+  /// HIGHEST`, default 5. This was **10** everywhere, which on iOS is actively
+  /// harmful and on Android is inert.
+  ///
+  /// **iOS** maps it straight onto URLSession: `BDPlugin.swift` does
+  /// `urlSessionDownloadTask.priority = 1 - Float(task.priority) / 10`. So 10
+  /// produced `0.0` — below `URLSessionTask.lowPriority` (0.25) — for a
```
781flyingdutchman, line 273: I am not sure where this `URLSessionTask.lowPriority = 0.25` comes from: the minimum allowed on iOS (and the one that gives it a very low priority) is 0.0 and a taskPriority of 10 gives you that. So, while you no longer use taskPriority of 10, this is not a bug

Issue #360, opened 2026-07-06
[Bug] AGP 9 + android.builtInKotlin=false: FlutterGemmaPlugin not compiled → GeneratedPluginRegistrant "cannot find symbol" (regression since 1.1.2 / #323)


ZuSe:
## Summary

Since **1.1.2** (the Built-in Kotlin migration, #323), the Android module only applies `kotlin-android` when `AGP major < 9`, otherwise relying on AGP 9's built-in Kotlin to compile the plugin's `.kt`. But AGP 9's built-in Kotlin is **opt-in** (`android.builtInKotlin=true`), and the Flutter AGP-9 migrator writes **`android.builtInKotlin=false`** by default. Apps that keep it `false` — which they must, if any other plugin applies KGP (see below) — get **no Kotlin compilation for `flutter_gemma`**, so `dev.flutterberlin.flutter_gemma.FlutterGemmaPlugin` is never emitted and the app's `GeneratedPluginRegistrant` fails to compile.

This is a regression: `1.1.0` applied KGP unconditionally and built fine on AGP 9.

## Error

```
/…/android/app/src/main/java/io/flutter/plugins/GeneratedPluginRegistrant.java:69: error: cannot find symbol
      flutterEngine.getPlugins().add(new dev.flutterberlin.flutter_gemma.FlutterGemmaPlugin());
                                                                        ^
  symbol:   class FlutterGemmaPlugin
  location: package dev.flutterberlin.flutter_gemma
1 error

> Task :app:compileReleaseJavaWithJavac FAILED
Execution failed for task ':app:compileReleaseJavaWithJavac'.
```

## Environment

| | |
|---|---|
| flutter_gemma | 1.2.1 |
| Flutter | 3.44.2 (stable) |
| AGP | 9.0.1 |
| Gradle | 9.1.0 |
| Kotlin | 2.3.20 |
| `android/gradle.properties` | `android.builtInKotlin=false`, `android.newDsl=false` (both written by the Flutter migrator) |

The app module deliberately does **not** apply `kotlin-android` (Kotlin comes from the Flutter Gradle plugin for the app module); every Kotlin **plugin** module compiles via its own KGP.

## Root cause

`flutter_gemma/android/build.gradle`:

```gradle
def agpMajor = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0] as int
if (agpMajor < 9) {
    apply plugin: 'kotlin-android'
}
```

[...]

PR #452, opened 2026-08-21
fix(android,ios): drop KGP from the Android plugins (#440); core iOS floor 16 → 15 (#441)

### LINKED ISSUE #440: Migrate plugins to built-in Kotlin
I am using `flutter_gemma` in my Flutter app.

Starting with Android Gradle Plugin (AGP) 9.0, support for applying the Kotlin Gradle Plugin (KGP) has been removed.

Because this plugin applies KGP, it causes a compilation error that prevents my app from building. Flutter [issue #181383](https://github.com/flutter/flutter/issues/181383) includes an example of the error.

Flutter has temporarily added support to allow KGP while apps and plugins migrate to AGP 9.0 or later, but this support will be removed in a future version of Flutter.

Please migrate this plugin to use built-in Kotlin to ensure your plugin users can successfully
build their apps in future versions of Flutter.

Here is the Flutter [migration guide for plugin authors](https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors).

### LINKED ISSUE #441: Minimum iOS version
Flutter's current SDK release minimum [iOS version 15](https://flutter.dev/blog/whats-new-in-flutter-3-47#:~:text=Prepping%20for%20the%20next%20wave%20of%20Apple%20updates).

However the `flutter_gemma` package sets the minimum required iOS version to iOS 16, which seems to be only needed for `flutter_gemma_mediapipe` module.

It would be great if only the `flutter_gemma_mediapipe` required iOS 16 with the rest of packages be iOS 15.

DenisovAV:
Closes #440. Closes #441. Both reported by @ekuleshov.

## #440 — the plugins applied KGP themselves

All three Android modules carried `ext.kotlin_version`, the KGP classpath, and a conditional `apply plugin: 'kotlin-android'`. Flutter's tooling now enumerates every plugin that does this:

```
WARNING: Your app uses the following plugins that apply Kotlin Gradle Plugin (KGP):
         flutter_gemma, flutter_gemma_builtin_ai, flutter_gemma_mediapipe
Future versions of Flutter will fail to build if your app uses plugins that apply KGP.
```

The precondition for dropping KGP outright is a `flutter: '>=3.44.0'` floor, which all three already declare. **Flutter's own Gradle plugin applies `kotlin-android` to any plugin subproject that does not** — `FlutterPluginUtils.kt:635`, `if (!hasKgpPlugin) pluginManager.apply("kotlin-android")` — so the Kotlin still compiles and the `kotlin {}` extension is still registered for the jvmTarget block.

`jvmTarget` moves from the deferred `pluginManager.withPlugin('kotlin-android') { kotlin { … } }` to a plain top-level `kotlin {}`. The old file's comment warned that a bare block "fails with Could not find method kotlin()" — that was true when the plugin decided for itself whether KGP was applied; it is not true now that Flutter always applies it.

This supersedes the two previous rounds in this area: #323 (apply KGP only on AGP < 9) and #360 (…unless built-in Kotlin is off). Both were guards on *when to apply KGP*. There is no longer anything to guard.

### Verified

A scratch app from `flutter create` on this SDK — **AGP 9.0.1, Gradle 9.1.0, `android.builtInKotlin=false`**, which is exactly #360's configuration — with all three packages as path deps:

| | build | Flutter's KGP warning |
|---|---|---|
| plugin sources at `main` | ✅ | lists **flutter_gemma, flutter_gemma_builtin_ai, flutter_gemma_mediapipe** |
| this branch | ✅ | *(none)* |

Both build today; the change is what the warning says it is — protection against a future Flutter that fails instead of warning.

Also built on **AGP 8.11.1** (the example, `--release`). Plugin Kotlin genuinely compiled in both: `mapping.txt` lists all three plugin classes as R8 input (`FlutterGemmaPlugin -> j4.b` etc.), and the AGP 9 debug dex carries them under their real names.

### One thing this does not fix

`android.builtInKotlin=true` still fails on AGP 9, and not because of us. Control: a bare `flutter create` app plus **`shared_preferences`** and nothing else fails identically —

```
Failed to apply plugin 'com.android.internal.library'.
> The 'org.jetbrains.kotlin.android' plugin is no longer required for Kotlin support since AGP 9.0.
```

— because Flutter applies KGP to the plugin subproject and AGP 9 rejects that when built-in Kotlin is on. The same plain app *without* plugins builds fine. So `builtInKotlin=true` is currently unusable for any Flutter app with plugins; `false` is what the template writes and what the migration guide targets, and that is the configuration verified above.

Separately worth knowing: `background_downloader` (a core dependency) applies KGP via `apply plugin: ['kotlin', 'android'].join('-')`. The id is assembled at runtime, so Flutter's scanner does not see it and it never appears in the warning list.

## #441 — core's iOS floor was MediaPipe's

Core declared `platform :ios, '16.0'`. Its entire iOS surface is a 37-line method channel — `FlutterMethodChannel`, `Bundle.main.path` — with no `@available` anywhere. The highest floor in its transitive native graph is `background_downloader`'s **14.0** (`large_file_handler` 12.0, `shared_preferences_foundation` 13.0).

The 16 was MediaPipe GenAI's, and MediaPipe has had its own package and its own podspec since the 1.0 split. Our own README said so — "Required for MediaPipe GenAI" — while the floor stayed on core.

`flutter_gemma_builtin_ai` carried the same borrowed 16 while gating **every** Foundation Models call behind `#available(iOS 26.0, macOS 26.0, *)`; leaving it at 16 would have reproduced the complaint for anyone using it. Both go to 15.0, in the podspec and the SPM manifest. `flutter_gemma_mediapipe` stays at 16.0 — it is the one that needs it.

15.0 rather than 14.0 to match where the Flutter SDK itself is going (3.47), and it is above every constraint above, so it is not a claim we have to walk back.

### Verified

A scratch app pinned to `IPHONEOS_DEPLOYMENT_TARGET = 15.0` with core + builtin_ai (no mediapipe), built for the simulator:

- **at `main`** → fails with the reporter's error:
  `The package product 'flutter-gemma' requires minimum platform version 16.0 for the iOS platform, but this target supports 15.0` (and the same for `flutter-gemma-builtin-ai`)
- **on this branch** → builds; `Info.plist` `MinimumOSVersion = 15.0`

Docs updated in six places — README, installation, troubleshooting — to state the floor per engine instead of prescribing 16 unconditionally. Two stale lines inside builtin_ai's own podspec and `Package.swift` that still said "iOS 16 floor" were corrected with it.

## Versions

`flutter_gemma` 1.6.4, `flutter_gemma_builtin_ai` 0.1.1. `flutter_gemma_mediapipe` stays 1.0.5 — bumped earlier and never published, so #440 folds into that entry. Podspec versions moved in lockstep.


ekuleshov:
@DenisovAV FYI.
Your AI's callout about `background_downloader` is not taking into account the whole picture:

```
def agpMajor = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0] as int
def isBuiltInKotlinFalse = project.hasProperty('builtInKotlin') && (project.property('builtInKotlin') == false || project.property('builtInKotlin') == 'false')
if (agpMajor < 9 || isBuiltInKotlinFalse) {
    apply plugin: ['kotlin', 'android'].join('-')
}
```

DenisovAV:
You are right, and thanks for taking the time. My AI got ahead of itself there and I did not check it carefully enough before posting — the line was quoted without the guard sitting directly above it, which is what gives it its meaning.

On AGP 9 that branch does not run: `agpMajor < 9` is false, and `isBuiltInKotlinFalse` is false as well, because the condition reads the property `builtInKotlin` while an app sets `android.builtInKotlin` — so `hasProperty` never finds it.

Checking it properly turned up something that inverts the point entirely. KGP **is** applied to `:background_downloader` on AGP 9 — by Flutter, not by the package. `FlutterPluginUtils.detectApplyingKotlinGradlePlugin` matches on literal text, cannot see the assembled id, concludes the module has no KGP, and applies `kotlin-android` itself. A live Gradle probe confirms `kgpApplied=true` on that project.

I tested the counterfactual, since the claim is only worth as much as the check behind it: replacing `['kotlin', 'android'].join('-')` with a literal `'kotlin-android'` — identical runtime behaviour — breaks the build on AGP 9 with `android.builtInKotlin=false`:

```
Could not find method kotlin() ... on project ':background_downloader'
```

With a literal id Flutter sees `hasKgpPlugin = true` and steps back, while the guard keeps the package's own apply from firing, so nothing applies KGP and its `kotlin {}` block has no extension to bind to.

So in that configuration the assembled id is load-bearing rather than a gap — the opposite of what my note implied. To be precise about the scope: on AGP 8 the guard does fire and the package applies KGP itself, so the token choice does not matter there; and with `android.builtInKotlin=true` the build fails either way, so it does not matter there either. It is specifically the AGP 9 + `builtInKotlin=false` lane — the one Flutter's own migrator writes — where it carries weight. Sorry for the noise, and thanks for the correction.

PR #468, opened 2026-08-27
fix(core): a web session right after install could not find the model — 1.7.2

DenisovAV:
Found by finally running the web litertlm suite end to end. All 7 tests failed
with **"No active inference model set"** over a model that had just installed
successfully.

## Narrowed in four steps, each measured

| step | result | rules out |
|---|---|---|
| `install()` returns, `getActiveModel()` works | ok | the install path |
| prefs after install hold **1 key of 4** | partial identity | — |
| four `setString` in a row read back fine | ok | `shared_preferences` |
| `setActiveModel` does `unawaited(_persist…)` | **found it** | — |

## The race

`setActiveModel` is `void` in `ModelFileManager`, so it cannot await the
persistence it starts — it can only fire it. Invisible until something reads the
result back, and on web something does: `WebModelSourceResolver.forActiveModel()`
builds a **fresh** `WebModelManager` and rehydrates the identity from prefs.
`createSession()` right after an install did exactly that, before the remaining
three writes had run. Every failure said *"after the test had completed"* — the
tail of the unawaited future.

This is **ordering, not durability**. `shared_preferences` updates its in-memory
cache synchronously inside `setString`; only the platform write is async
(verified in the package source). Once the writes have *run*, any reader in the
isolate sees them — so no API migration is warranted. `SharedPreferencesAsync`
solves the multi-isolate problem, which is not this one.

## The fix

A **library-level** write barrier — library-level precisely because the reader is
a different instance — that chains the writes and is awaited by `_doInit` before
it reads prefs. Same single-flight idiom the file already uses for `_initFuture`.

`unawaited` is gone from the three persist paths. Its own dartdoc says it conveys
*"the intention of deliberately not waiting"* and that an error there *"was likely
a mistake to not await"*; here a later reader depends on the result.

Also removed: the `catch` in each persist method that logged and continued. With
`gemmaLog` compiled out of release, an identity that never got written said
nothing at all. Failures now reach the barrier, which records them once — and the
resolver's error names the cause instead of telling the caller to install a model
they already installed.

## Verified

Same suite, same arrangement either side of the change: **7 failures before,
0 after**. `flutter analyze` 0 errors/warnings. Core **1.6.6** with both podspecs.

## Correction

An earlier push of this branch used `git add -A` and swept in unrelated
in-progress work from the working tree — a vendored sentencepiece package and the
edits around it, about 6400 lines. That is also where the "two pre-existing red
tests on main" I reported came from: `siglip_tokenizer_parity_test.dart` arrived
with that work and hardcodes an absolute path to a different project on the
machine, and `ort_client_test.dart` was broken by the swept-in `ort_client.dart`
edit. Neither was on main and neither was pre-existing — I had just committed
them.

The branch is now seven files, staged individually. The swept-in work is back in
the working tree, untouched.


DenisovAV:
Rebased onto `main` and retargeted: this now releases core **1.7.2**, not 1.7.6 — 1.7.0 and 1.7.1 shipped while the branch sat. All five conflicts were version lines (pubspec, both podspecs, `CLAUDE.md`, the CHANGELOG heading); the entry moves above main's 1.7.1 section rather than replacing it.

Two defects in what this branch added around the barrier, both found on review and fixed in `5ebcacb3`:

**The cause never reached the caller.** `'after install (\$why)'` escapes the dollar inside a single-quoted string, so the one new sentence rendered literally as `after install ($why)`. Valid Dart, `analyze` silent — and the whole point of the message lost.

**The failure marker went stale.** It is a library-level field written once and never cleared, so a single failed write kept explaining every later "No active inference model set" for the life of the isolate, including after a different model had installed fine. Writes are chained through the barrier, so the newest one is the current answer; it now clears on success.

Also a dangling doc reference to `[_persistFailure]`, which is not a symbol here.

**Test added.** `test/web/web_identity_write_barrier_test.dart` asserts that a manager built right after `setActiveModel` — the issue's shape, since `WebModelSourceResolver.forActiveModel()` builds a fresh one — sees all four identity keys. It asserts on prefs rather than on `activeInferenceModel`, which also needs the model file present; that was my first attempt and it failed even with the barrier in place.

It is `@TestOn('browser')` because `WebModelManager` pulls in `dart:js_interop` and cannot load on the VM, so `tool/test_all.sh` does not run it:

```
flutter test test/web/web_identity_write_barrier_test.dart --platform chrome
```

Commenting out the `await _identityWriteBarrier` in `_doInit` turns it red on three consecutive runs, so it pins the barrier rather than a race.

VM suite 701 pass, `flutter analyze packages/` clean.

DenisovAV:
Replaced the write barrier with the reorder that removes the race outright, after design review. The barrier, the failure marker and `webIdentityWriteFailure` are gone; what ships is 15 lines in one file.

**The barrier's premise was wrong.** It was library-level state "precisely because the reader is a different instance" — but instance identity never mattered for the data. Verified in `shared_preferences` 2.5.5:

```dart
static Future<SharedPreferences> getInstance() async {
  if (_completer == null) { ... }        // static — one instance per isolate

Future<bool> _setValue(String valueType, String key, Object value) {   // NOT async
  _preferenceCache[key] = value;         // cache written synchronously
  return _store.setValue(valueType, prefixedKey, value);
}
```

Every `WebModelManager` shares that one `_preferenceCache`, and the cache is written before the call can yield. So a run of `setString`s with **no `await` between them** is already atomic with respect to every reader in the isolate. Awaiting each in turn is the only thing that made the writes observable half-done.

Issuing the four together closes both variants structurally instead of ordering around them:

- the reported one, a first install — the reader sees a partial identity (measured: one key of four) and throws "No active inference model set" over a model that just installed;
- the unreported and worse one, a **re-install** — prefs hold the previous identity until each key is overwritten, so a reader catching the write half-done gets the new `modelType` with the old `filename`. That is well-formed, passes `isInstalled`, and loads the wrong weights with nothing thrown.

The barrier could not have caught the second case anyway: its diagnosis is read only on the `active == null` branch of `resolveActiveInferenceModel`, and a mixed identity has all four keys present, so the restore rebuilds a wrong spec and takes the success path where the marker is never consulted.

Applied to the embedding and STT identities too — same shape, same exposure.

**Tests.** `test/web/web_identity_write_atomicity_test.dart` covers the first install and the re-install. `@TestOn('browser')`, because `WebModelManager` pulls in `dart:js_interop` and cannot load on the VM, so `tool/test_all.sh` does not run it:

```
flutter test test/web/web_identity_write_atomicity_test.dart --platform chrome
```

Restoring the sequential awaits turns it red on three consecutive runs.

**Deliberately not in this patch,** left for 2.0 where `setActiveModel` returning `Future<void>` is already planned and makes this bug class structurally impossible: sharing one `WebModelManager` rather than building a throwaway per `createModel`, and the `RuntimeConfig` asymmetry that has web engines resolve the model themselves while native engines are handed a resolved path. Both are real, both are bigger than a patch release, and building barrier machinery now would only be deleted there.

VM suite 701 pass, `flutter analyze packages/` clean.

Issue #454, opened 2026-08-24
Proposal: optional per-repo litertlm_manifest.json for custom .litertlm models (variant choice, backend, capabilities, session defaults)


john-rocky:
Thanks for flutter_gemma — the FFI engine and the pluggable engine registry have made it the default way Flutter apps run `.litertlm` models. I maintain the conversion pipeline behind a number of the files people load through it ([hf-to-litertlm](https://github.com/john-rocky/hf-to-litertlm); models hosted under `litert-community` and `mlboydaisuke` on Hugging Face), including several on your supported list (SmolLM3, FastVLM, Qwen2-VL) and several outside it (LFM2.5, Qwen3.5, MiniCPM).

For a model outside the built-in list, the caller has to know things the package cannot. Which of a repo's files to pick: [litert-community/LFM2.5-1.2B-Instruct](https://huggingface.co/litert-community/LFM2.5-1.2B-Instruct) ships an int4 that cannot create a GPU engine next to an `_int4_gpu` re-export that can. Which backend is fastest for the device class. Vision/audio/thinking capability. And session defaults — a reasoning model needs max output tokens ≥ 2048 or it is cut off inside its `<think>` chain. Today I put all of that in per-model README prose, which no package can read.

I've started publishing it machine-readable instead: a repo-level `litertlm_manifest.json` covering exactly the deployment layer — variant selection, per-platform backend recommendations, requirements, capabilities, and measured performance rows that carry their conditions (runtime version, token counts, cache mode, run count). It duplicates nothing the bundle header already carries; bundle-derived fields are generated by reading the published file's header with two HTTP range requests.

- Spec and layering rule: [manifest/SCHEMA.md](https://github.com/john-rocky/hf-to-litertlm/blob/main/manifest/SCHEMA.md)
- Finished examples: [LFM2.5-1.2B-Instruct](https://github.com/john-rocky/hf-to-litertlm/blob/main/manifest/examples/litert-community__LFM2.5-1.2B-Instruct.json), [Qwen3-4B-Thinking-2507](https://github.com/john-rocky/hf-to-litertlm/blob/main/manifest/examples/litert-community__Qwen3-4B-Thinking-2507.json)
- Generator: [manifest/make_manifest.py](https://github.com/john-rocky/hf-to-litertlm/blob/main/manifest/make_manifest.py)

[...]

Issue #318, opened 2026-06-15
Node number 1149 (DYNAMIC_UPDATE_SLICE) failed to prepare


GZHarry:
Hello! I wanted to ask, the latest updated speculative decoding model: gemma-4-E2B-it.litertlm, cannot infer images or text on Android using flutter_gemma: 0.16.5. Could you help me see how to adjust it, thanks! (Model download link: https://huggingface.co/litert-community/gemma-4-E2B-it-litert-lm), the error is as follows:

I/tflite  (25604): Initialized TensorFlow Lite runtime.
I/flutter (25604): 文件保存路径已存在
I/flutter (25604): ℹ️  Model already installed: gemma-4-E2B-it.litertlm (skipping download)
I/flutter (25604): ✅ Set active inference model: gemma-4-E2B-it
I/flutter (25604): ✅ Inference model installed and set as active: gemma-4-E2B-it
I/BufferQueueConsumer(25604): [](id:640400000000,api:0,p:-1,c:25604) connect(): controlledByApp=true
I/BufferQueueConsumer(25604): [](id:640400000001,api:0,p:-1,c:25604) connect(): controlledByApp=true
I/BufferQueueConsumer(25604): [](id:640400000002,api:0,p:-1,c:25604) connect(): controlledByApp=true
E/CameraManagerGlobal(25604): Camera 3 is not available. Ignore physical camera status change
W/CameraManagerGlobal(25604): ignore the torch status update of camera: 2
W/CameraManagerGlobal(25604): ignore the torch status update of camera: 3
I/Camera  (25604): startPreview
D/TTS     (25604): Utterance ID has started: c1ce1628-3a38-4e61-bd28-65fda5c58433
D/TTS     (25604): Utterance ID has completed: c1ce1628-3a38-4e61-bd28-65fda5c58433
I/BufferQueueProducer(25604): [SurfaceTexture-0-25604-0](id:640400000000,api:4,p:885,c:25604) connect(): api=4 producerControlledByApp=true
I/BufferQueueProducer(25604): [ImageReader-1920x1080f100m1-25604-0](id:640400000001,api:4,p:885,c:25604) connect(): api=4 producerControlledByApp=false
I/Camera  (25604): CameraCaptureSession onConfigured
I/Camera  (25604): Updating builder settings
D/Camera  (25604): Updating builder with feature: ExposureLockFeature
D/Camera  (25604): Updating builder with feature: ExposurePointFeature
[...]

PR #469, opened 2026-08-27
feat(litertlm): LitertlmManifestResolver — manifest-driven Hugging Face install (#454)

john-rocky:
The resolver half of #454, built on the seam from #461: `flutter_gemma_litertlm` gains **`LitertlmManifestResolver`**, a `HuggingFaceResolver` for repos that ship a [`litertlm_manifest.json`](https://github.com/john-rocky/hf-to-litertlm/blob/main/manifest/SCHEMA.md) deployment manifest. Register it via `initialize(huggingFaceResolvers:)`, and `resolveHuggingFace(repo, fileType: ModelFileType.litertlm)` returns the right `.litertlm` file for the platform plus overridable runtime defaults — no hardcoded filenames. The README gains the end-to-end snippet.

## What this adds (`lib/src/manifest/`)

- **`LitertlmManifestResolver`** — claims exactly `ModelFileType.litertlm`, never a null hint (the #461 contract). Fetches `…/resolve/<revision>/litertlm_manifest.json` (Bearer token when given; a constructor `revision:` pins both the manifest and the returned file URL, so installing via `fromNetwork(r.url)` keeps the pin). Selection is the manifest spec's v0.1 algorithm, vendored from the reference Dart reader and kept logic-identical, so every integration resolves a repo to the same file.
- **IO** — an injectable `ManifestFetch`; the defaults are `dart:io`'s `HttpClient` natively and the browser's `fetch` on web. No new dependencies. Both follow Hugging Face's 307 redirect on `/resolve/` paths.
- **Errors with guidance** — 404: the repo ships no manifest, install with `fromHuggingFace(repo, file:)`. 401/403: token guidance — measured Hugging Face behaviour is 401 (not 404) for unauthenticated gated, private, *and nonexistent* repos, since repo existence isn't revealed. A non-0.1.x manifest is refused with a `FormatException` rather than half-parsed.

## Mapping (the decisions to veto)

- `maxTokens` ← `model.context_length` (the engine still clamps downstream, #318); `supportImage`/`supportAudio` ← `capabilities`; `isThinking` ← `capabilities.thinking.declared` — an object in the spec; the exact channel markers stay available in the manifest for a future `createSession(defaults:)`.
- `minOutputTokens` ← `session_defaults.max_output_tokens_min`. `session_defaults` is schema-open, so the key is read by name as an optional int and junk types are ignored. I've also declared the in-the-wild keys additively in the spec's 0.1 line ([schema commit](https://github.com/john-rocky/hf-to-litertlm/commit/c27de1e)), so a wrong type now fails validation at manifest-generation time.
- `preferredBackend` ← the resolved backend. An explicit caller hint follows the spec's resolution rule — a *filter*, not a preference: only variants verified on it compete and the result keeps exactly that backend; when no variant lists it, the hint is dropped with a log line and the manifest's own choice comes back, never a silent claim of the requested backend. A name outside `{cpu, gpu, npu}` maps to null (SDK default) with a log line.
- `modelType` — conservative, **null over a wrong guess**: on iOS `.litertlm` chats are formatted manually by `ModelType`, so a wrong family breaks conversations. Family words match only the curated `base_model`/`display_name` ids — the free-prose `architecture` never maps by itself (granite-docling's "Llama-architecture granite decoder" must not become `ModelType.llama`) — plus exactly two machine-precise class tokens, `Qwen3ForCausalLM`/`Qwen2ForCausalLM`, which catch finetunes whose ids hide the family. Qwen3.5 deliberately maps to `qwen`, not `qwen3`: same ChatML handling, but `qwen3` also injects ` /no_think` into user turns, which Qwen3.5 would read as literal text. Over today's catalog: `deepSeek` 1 (the R1 distill — `deepseek` outranks the `qwen` in its id), `qwen3` 2, `qwen` 6, null 12.
- `notes` ← `platform_notes` + `known_issues` + the curated `session_defaults.notes` string. Sampler hints (`temperature`/`top_k`/`top_p`, present in one shipped manifest) have no `ModelRuntimeDefaults` home yet — they wait for `createSession(defaults:)`.
- v1 resolves on **platform only**; `device_class` stays advisory (free strings, no defined vocabulary, no detection in core) — per the #454 agreement.

## Testing

- 52 contract/unit tests: fake-fetch mapping, registry probe, URL/revision/per-segment encoding, failure modes, the full `modelType` table, and a vendored-reader parity group mirroring the reference reader's own suite (explicit-backend filter, null over substitution, unverified recommendations ignored, deviceClass fallback notes, 0.1.1 `declaredChannels`).
- A regression sweep over a snapshot of **every shipped manifest** (21 repos / 35 variants as of 2026-08-27; `fixtures/README.md` has the regeneration one-liner): 672 platform × backend-hint combinations hold the invariants — the resolved backend is always in the variant's *verified* `backends` list, a hint some variant is verified on is honoured exactly, identity fields come from the chosen variant, model-level fields land regardless of variant choice. Golden selections are pinned to the reference readers' output, and dataset-shape counts keep the mapping honest: 21 of the 35 variants carry no `recommended`, so the `default_backend` fallback is the resolver's main path, not an edge.
- Live-verified against Hugging Face: platform selections, the revision pin, 307 following, and the 401/404 semantics above.
- Full package suite: 94 tests green; `flutter analyze` reports nothing for the new code.

## Not in this PR

- Acting on the sampler hints (waits for `createSession(defaults:)`).
- `device_class` resolution, and post-download sha256/size verification (carried as advisory, per #461).
- Exporting the raw manifest-reader types — `LitertlmManifestResolver` plus the fetch seam are the public surface; happy to widen it if apps want the `measured[]` rows.

Version is bumped 1.5.2 → 1.6.0 on the branch with a one-line CHANGELOG entry — renumber as you prefer. The `flutter_gemma` constraint is untouched: it needs whatever release carries #461, and that number is yours.

### DIFF
--- a/packages/flutter_gemma_litertlm/CHANGELOG.md
+++ b/packages/flutter_gemma_litertlm/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 1.6.0
+- Add `LitertlmManifestResolver`: manifest-driven Hugging Face install for repos shipping `litertlm_manifest.json` (#454).
+
 ## 1.5.3
 - Fix a rebuilt native library not reaching the build.
 
--- a/packages/flutter_gemma_litertlm/README.md
+++ b/packages/flutter_gemma_litertlm/README.md
@@ -27,6 +27,42 @@ await FlutterGemma.initialize(
 other engines (e.g. `MediaPipeEngine` from `flutter_gemma_mediapipe`) if your app
 uses both formats.
 
+## Install from a Hugging Face repo (`litertlm_manifest.json`)
+
+Repos that ship a
+[`litertlm_manifest.json`](https://github.com/john-rocky/hf-to-litertlm/blob/main/manifest/SCHEMA.md)
+deployment manifest describe every `.litertlm` file they contain — which
+backends each is verified on, which file a given platform should pick, sha256/
+size identity, and session guidance. `LitertlmManifestResolver` reads it so an
+app installs "the right file for this device" without hardcoding filenames:
+
+```dart
+await FlutterGemma.initialize(
+  inferenceEngines: [LiteRtLmEngine()],
+  huggingFaceResolvers: [const LitertlmManifestResolver()],
+);
+
+final r = await FlutterGemma.resolveHuggingFace(
+    'litert-community/Qwen3-4B-Thinking-2507',
+    fileType: ModelFileType.litertlm);
+await FlutterGemma.installModel(
+      modelType: r.modelType ?? ModelType.general,
+      fileType: r.fileType,
+    )
+    .fromNetwork(r.url) // authoritative: carries the resolver's revision pin
+    .install();
+final model = await FlutterGemma.getActiveModel(defaults: r.runtime);
+final session = await model.createSession(
+  enableThinking: r.runtime.isThinking ?? false,
+  maxOutputTokens: r.runtime.minOutputTokens, // a floor — see the field docs
+);
+```
+
+Everything the manifest returns is an overridable default (explicit argument >
+manifest > SDK default); `r.notes` carries platform caveats and known issues
+worth surfacing to developers. Repos without a manifest keep working through
+`installModel(...).fromHuggingFace(repo, file: ...)`.
+
 ## Embeddings
 
 ```dart
--- a/packages/flutter_gemma_litertlm/README.md
+++ b/packages/flutter_gemma_litertlm/README.md
@@ -37,6 +37,8 @@ size identity, and session guidance. `LitertlmManifestResolver` reads it so an
 app installs "the right file for this device" without hardcoding filenames:
 
 ```dart
+import 'dart:math' show max;
+
 await FlutterGemma.initialize(
   inferenceEngines: [LiteRtLmEngine()],
   huggingFaceResolvers: [const LitertlmManifestResolver()],
@@ -54,7 +56,9 @@ await FlutterGemma.installModel(
 final model = await FlutterGemma.getActiveModel(defaults: r.runtime);
 final session = await model.createSession(
   enableThinking: r.runtime.isThinking ?? false,
-  maxOutputTokens: r.runtime.minOutputTokens, // a floor — see the field docs
+  // minOutputTokens is a floor, not a cap: keep the app's own budget unless
+  // the manifest asks for more.
+  maxOutputTokens: max(1024, r.runtime.minOutputTokens ?? 0),
 );
 ```
 
--- a/packages/flutter_gemma_litertlm/lib/flutter_gemma_litertlm.dart
+++ b/packages/flutter_gemma_litertlm/lib/flutter_gemma_litertlm.dart
@@ -26,3 +26,12 @@ export 'src/ffi/litert_bindings_stub.dart'
 // the LiteRT.js-backed `WebEmbeddingModel` directly.
 export 'src/embedding/litert_embedding_backend_web.dart'
     if (dart.library.ffi) 'src/embedding/litert_embedding_backend.dart';
+
+// Hugging Face resolver for repos that ship a `litertlm_manifest.json`
+// deployment manifest. Register with
+// `FlutterGemma.initialize(huggingFaceResolvers: [LitertlmManifestResolver()])`
+// and drive it via `FlutterGemma.resolveHuggingFace(repo,
+// fileType: ModelFileType.litertlm)`. All six platforms (its IO arm picks
+// dart:io or browser fetch internally).
+export 'src/manifest/litertlm_manifest_resolver.dart'
+    show LitertlmManifestResolver, ManifestFetch, ManifestFetchException;
--- a/packages/flutter_gemma_litertlm/lib/src/manifest/litertlm_manifest.dart
+++ b/packages/flutter_gemma_litertlm/lib/src/manifest/litertlm_manifest.dart
@@ -1,0 +1,381 @@
+/// Reader for `litertlm_manifest.json` — the deployment manifest shipped at
+/// the root of `.litertlm` model repos on Hugging Face.
+///
+/// Vendored from the reference reader at
+/// https://github.com/john-rocky/hf-to-litertlm/tree/main/readers/dart (same
+/// author); the resolution algorithm here and there must stay identical, so a
+/// repo resolves to the same file no matter which integration reads it.
+/// Spec: https://github.com/john-rocky/hf-to-litertlm/blob/main/manifest/SCHEMA.md
+///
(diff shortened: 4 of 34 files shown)


DenisovAV:
This is great — I ran a thorough multi-agent pass and verified everything against the code, and the resolver is solid and faithful to the seam. `canResolve` claims exactly `ModelFileType.litertlm`, no seam-type duplication, the 0.1.x gate and per-segment URL/revision pinning are correct, `dart:io` stays out of the web graph with no conditional-import drift, and the coverage is genuinely thorough — the 21-repo/35-variant sweep and the `mapModelType` suite discriminate real branches, not tautologies. The happy path (fetch a real manifest → select the variant → map to defaults) is well-tested and clean. Nice work.

Nothing here is a blocker. A couple of small cleanups if you're in there, then a few optional notes:

**Worth a quick fix (both tiny, both only bite malformed manifests, but they break the documented `FormatException` contract):**

1. `manifest.resolve()` and the `rawModel['base_model']/['architecture']` reads run outside the parse `try/catch`, and `.cast<String>()` on `platform_notes`/`known_issues`/`backends` is lazy — so a bad element (`"platform_notes": ["ok", 42]`, or `"base_model": 3`) slips past `fromJson` and later throws a raw `TypeError` instead of the `FormatException` the docs promise. `.cast<String>().toList()` at the three sites + moving the `base_model`/`architecture` reads inside the `try` closes it.
2. `manifest_fetch_io` decodes the body before checking the status, so a non-200 whose body fails to UTF-8-decode loses its `statusCode` and the resolver can't tell 404-from-network-error. Checking status before reading the body (or `allowMalformed: true`) fixes it. (io also accepts only `200` while web accepts any 2xx — worth aligning while you're there.)

**Optional / no impact today, noting for the record:**

- `Capabilities.fromJson(null)` maps a manifest with no `capabilities` block to `false` rather than `null`, collapsing the seam's `bool?` "silent vs. explicitly-unsupported" distinction. Every current consumer does `?? false`, so it's harmless now — just a deliberate call for later.
- A dropped `preferredBackend` (no variant verifies it) is only reported via `gemmaLog` (release-stripped); appending it to `ResolvedHfModel.notes` would give release builds a trace of the downgrade.
- `max_output_tokens_min` accepts a negative int and silently drops a wrong-typed one — worth a small range/type guard.
- The README/doc snippet passes the `minOutputTokens` floor straight in as `maxOutputTokens`; the caveat comment helps, but `max(minOutputTokens, appDefault)` models the "floor, not a cap" intent better.
- Quick question, not a bug: with no `device_class` match, `resolve()` uses the first platform recommendation (and notes it), and on the SDK path `deviceClass` is always null — so a device-class-scoped rec applies to every device on that platform. Deliberate v1 (platform-only), just confirming that's the intended semantics vs. falling to `default_backend`.

On versioning: the `flutter_gemma:` floor needs to point at the core version that ships the seam (it landed on `main` after `1.6.5` without a bump) — we're publishing that core bump separately, so nothing for you to do here beyond moving the floor once it's out.

Thanks for building this out so cleanly on the seam — really nice contribution.

john-rocky:
Thanks for the careful pass — and for checking the seam mapping and the sweep against the code rather than the description. Both quick fixes and the four optional notes are in one follow-up commit on the branch ([63a2576](https://github.com/john-rocky/flutter_gemma/commit/63a25763b3c04354b72e498bc0d8418c895982ab)); the package suite goes 94 → 104, all green. On `device_class`: platform-only is deliberate for v1, and on today's shipped manifests it picks the same file and backend that falling to `default_backend` would — details at the end.

**1. FormatException contract.** `.cast<String>().toList()` at the three reader sites, and the `base_model` / `architecture` / `capabilities` reads moved inside the parse `try`. A wrong-typed element or field is now a `FormatException` naming the repo, with tests for each of the `platform_notes` / `known_issues` / `backends` elements and for `base_model` / `architecture`. After that, nothing `resolve()` touches is lazily typed, so it stays outside the `try`.

**2. Fetch.** `manifest_fetch_io` reads the status first; any 2xx is success, matching the web arm's `Response.ok`; a non-2xx body is drained, never decoded; a 2xx body decodes with `allowMalformed: true`, which is what the browser's `Response.text()` does, so both arms hand the parser the same body for the same bytes. A loopback `HttpServer` test pins it: a 404 whose body is not UTF-8 reports 404, 201/204 succeed, the 307 is followed (and one that lands on a 404 reports 404), and a refused connection is status-less.

**Optional notes, all taken:**

- No `capabilities` block → `supportImage` / `supportAudio` / `isThinking` are `null`. Decided at the seam (`rawModel['capabilities'] != null`), so the vendored reader stays identical to the reference; a present block still yields bools, with omitted keys `false`. The converter derives the block from the bundle header, so "absent" means the manifest never looked — the seam's "silent". All 21 shipped manifests carry the block, so no shipped result changes.
- A dropped `preferredBackend` is now the last entry of `notes`: `No variant of "<repo>" is verified on the requested npu backend; resolved without that hint (gpu).` The sweep asserts it on every drop and its absence on every honoured hint.
- `max_output_tokens_min` is taken only as an `int >= 1` (the schema's `integer, minimum: 1`); anything else is logged and ignored — an advisory knob shouldn't fail the whole manifest.
- README and the class doc now show `maxOutputTokens: max(1024, r.runtime.minOutputTokens ?? 0)`. The `resolveHuggingFace` doc in core still has the old form; I left core untouched here, and the same two-line change applies there whenever it suits you.

**`device_class`.** Yes, deliberate: v1 resolves on platform only, because `device_class` is a free string with no vocabulary in the 0.1 schema (`"type": "string"`) and flutter_gemma has no device-class detection to feed one. With no class requested, the reader takes a class-scoped row as the platform's recommendation rather than falling to `default_backend`, because the row is the manifest's verified-fastest statement for that platform, while `default_backend` is the converter's unranked default with no measurement behind it. One detail on the note: the "(no X entry; using Y)" text only fires when a class was requested and missed; on the SDK path nothing is requested, so the reason reads "recommended for android: …" with no class.

Measured on the shipped data: seven class-scoped rows exist, all `android`, in the three LFM2.5-1.2B repos; on the SDK path each resolves to the int4/cpu "midrange" row — the same file and backend the `default_backend` fallback returns — so the two semantics don't diverge on anything shipped today. A class-aware pick (`flagship` → int8_gpu) becomes reachable once the spec defines a vocabulary and core can pass one; the reader already ranks a class match above a platform match. If you'd rather the seam fell to `default_backend` whenever the only platform rows are class-scoped, that's a small change on my side — your call.

On versioning: understood — I'll move the `flutter_gemma:` floor as soon as the core release with the seam is out.

Thanks again for the review — it made the malformed-manifest paths a good deal more honest.

Issue #457, opened 2026-08-24
macOS Native Dylib Bundling Fails - post_install hook never executes


srinjoychinargoswami:
# macOS Native Dylib Bundling - post_install Hook Never Executes (Follow-up to #255)

## Related Issue
This is a follow-up to #255, which fixed the dylib search path for 0.14.4+. However, an architectural problem remains.

## Environment
- **Flutter Version:** 3.24.0+
- **flutter_gemma:** 1.6.0+
- **flutter_gemma_litertlm:** 1.5.0+
- **macOS:** Sequoia 15.x
- **Machine:** Apple M4

## What #255 Fixed

Issue #255 corrected the search path for companion dylibs. The 0.14.4 release changed the `macos/Podfile` snippet to look in the Native Assets cache:

```bash
${HOME}/Library/Caches/flutter_gemma/native/macos_arm64/
```

Instead of the non-existent prebuilt path. This was a good fix, but it only addressed **WHERE to find the dylibs**, not **WHETHER the bundling hook runs at all**.

## Remaining Problem

Even with 0.14.4+, the `post_install` hook in `macos/Podfile` **never executes**, causing the following runtime error:

Failed to load dynamic library 'LiteRtLm.framework/LiteRtLm':
dlopen(LiteRtLm.framework/LiteRtLm, 0x0001):
Library not loaded: @rpath/libGemmaModelConstraintProvider.dylib

### Root Cause

**The post_install hook never runs because Flutter macOS projects have zero CocoaPods dependencies by default.**

CocoaPods only executes `post_install` hooks when it has pods to process. With no external pod dependencies, the hook is silently skipped entirely.

Evidence from `pod install`:
```bash
$ pod install
Analyzing dependencies
Downloading dependencies
Generating Pods project
Integrating client projects
Pod installation complete! There are 0 dependencies from the Podfile and 0 total pods installed.
[!] The Podfile does not contain any dependencies.
```

The hook that should:
1. Copy dylibs from `~/Library/Caches/flutter_gemma/native/macos_arm64/` to `Contents/Frameworks/`
2. Patch `LiteRtLm.dylib`'s `LC_LOAD_DYLIB` references
3. Code-sign the dylibs

**...never runs at all.**

## Why The Current Documentation Fails

The README directs developers to add the post_install hook to `macos/Podfile` and run `pod install`. However:

1. The hook assumes it will execute during pod integration
2. It only executes if CocoaPods has work to do
[...]

PR #479, opened 2026-09-01
test(litertlm): offline manifest sweep on the live catalog + opt-in Hugging Face check (#454)

john-rocky:
The two pieces from #454, split the way you asked: the offline cross-product in the default suite, the live leg behind an opt-in.

**Offline (default `flutter test`).** The fixture snapshot moves to today's catalog: 28 repos / 51 variants, up from 21 / 35. 12 of the original manifests were reshipped since 08-27, and 7 repos are new. `fixtures/reference_goldens.json` is new: the hf-to-litertlm reference reader's selection for every repo × platform × backend hint, 896 rows, one per line. The sweep now pins every combination to it instead of eight hand-picked rows. The invariants and the distribution counts (floors, thinking, modelType) stay, updated to the new data, so a regenerated snapshot fails loudly rather than slide through. The README carries the regeneration recipe, including the goldens dump.

The registration contract — engine-carried auto-registration, an explicit resolver winning the equal-priority tie (and `resolveHuggingFace` reaching it), const dedup — is core's, so its test lives there: `flutter_gemma/test/core/registry/resolver_registration_test.dart` drives `FlutterGemma.initialize()` headless with a fake engine and const fake resolvers. The `SharedPreferences.setMockInitialValues({})` note is in its header. This package keeps its own half in `test/manifest/engine_carried_resolver_test.dart`: `FlutterGemma.engineHuggingFaceResolvers([LiteRtLmEngine()])` yields the canonical `const LitertlmManifestResolver()` — no `initialize()`, no mock. The only dependency change is `shared_preferences` in this package's dev_dependencies, for the live end-to-end test below, which still drives `initialize()`.

**Live (opt-in).** `live_hugging_face_test.dart` skips unless `FLUTTER_GEMMA_LIVE_HF` is set. With it, against the real Hugging Face:

- every fixture repo still serves the committed manifest through the published default fetcher (drift fails with the regeneration pointer);
- no repo outside the snapshot ships a manifest;
- every variant's `sha256` / `size_bytes` equal the repo's LFS metadata;
- every URL the resolver can build answers 200;
- `FlutterGemma.resolveHuggingFace` resolves through the engine-carried resolver, end to end.

The `HttpOverrides.global = null` note is in that file's header. `.github/workflows/live-manifests.yml` runs it weekly (Monday 06:23 UTC) and on `workflow_dispatch`, with no push or pull_request trigger. `HF_TOKEN` is optional.

Verified locally on Flutter 3.47.1 stable: `flutter test` — core 701 passed, this package 106 passed / 5 skipped (the live group). The live leg with the variable set: 5/5 in 64 s at the 09-02 snapshot; on 09-03 its two catalog checks report drift (3 manifests reshipped, 2 new repos), to be refreshed in a follow-up.

No version bump or changelog line: nothing here ships. Still not covered, as before: the install-time `ModelType.general` merge and `InferenceInstallation.notes` need an app run.


DenisovAV:
This is the split I asked for, done more carefully than I asked for it. Two things I want to name before the one request.

**The goldens change is the real upgrade here.** Pinning all 896 reference rows instead of eight hand-picked ones turns the sweep from "these examples still work" into "the resolver agrees with the reference reader everywhere". With 12 manifests reshipped under it since 08-27, that is the difference between a suite that would have noticed and one that would not.

**I checked the two properties that matter for our CI, rather than taking the description's word:**

- `live_hugging_face_test.dart` guards on `FLUTTER_GEMMA_LIVE_HF` with a group-level `skip:`, so a default `flutter test` skips it.
- `live-manifests.yml` declares only `schedule` and `workflow_dispatch` — no `push`, no `pull_request`. Exactly the semantics I wanted: its failure means "the catalog moved", never "this PR broke something".

I also measured the archive impact, since `test/` ships to pub.dev: the fixtures are 283 KB raw but **44 KB compressed** against a 338 KB package. That is fine — raising it only so you know it was considered and is not a reason to trim anything.

## One request: move `resolver_registration_test.dart` to core

The `shared_preferences` dev-dependency costs consumers nothing (core already depends on it directly, and dev-dependencies are not transitive), so this is not about the dependency itself. It is a symptom worth following: the test needs a storage mock because it drives `FlutterGemma.initialize()`, and what it asserts — registration order, the explicit resolver winning the equal-priority tie, `const` dedup — is **core's** contract, not this package's.

Core is already set up for it: `test/core/registry/hugging_face_resolver_registry_test.dart` is the sibling test, `setMockInitialValues` is already used in three of its test files, and `shared_preferences` is in its regular `dependencies`. Nothing new is needed there.

Two things improve in the move:

1. **In core, use a fake engine** — a small class implementing `InferenceEngineProvider` + `HuggingFaceResolverSource`. That tests the contract rather than one implementation, and core cannot depend on `flutter_gemma_litertlm` anyway (the dependency is one-directional).
2. **This package keeps the one question that is genuinely its own** — does `LiteRtLmEngine` contribute `LitertlmManifestResolver`? — answered by `FlutterGemma.engineHuggingFaceResolvers([LiteRtLmEngine()])`. That is the `@visibleForTesting` seam #471 added with the comment "extracted so the derivation is unit-testable without a full `initialize`", which is exactly this case. No `initialize()`, no storage mock, and the dev-dependency comes back out.

Your `SharedPreferences.setMockInitialValues({})` and `HttpOverrides.global = null` notes should travel with the tests that need them — the first to core, the second staying here.

## One logistics note

#478 also touches `packages/flutter_gemma_litertlm/pubspec.yaml` (a version bump for the #457 fix). Different region, so no textual conflict expected, but whichever lands second may want a rebase.

Everything else reads good to me. Thanks for the regeneration recipe in the fixtures README — that is the part that keeps this maintainable after both of us have forgotten how it was built.

john-rocky:
Thanks for the careful read — and for measuring the archive; good to have that number on record.

The move is in [3364201](https://github.com/DenisovAV/flutter_gemma/pull/479/commits/336420116840a3abef77a94b9ed70248c0b30249), along the two lines you drew: core holds the contract, this package keeps only the handover question. `analyze-and-test` is green on the push.

- `flutter_gemma/test/core/registry/resolver_registration_test.dart`: a fake engine (`InferenceEngineProvider` + `HuggingFaceResolverSource`) and const fake resolvers through the full `initialize()` path — auto-registration, the explicit list first with `resolveHuggingFace` reaching the explicit one, const dedup. The `setMockInitialValues` note travelled with it.
- `flutter_gemma_litertlm/test/manifest/engine_carried_resolver_test.dart`: `FlutterGemma.engineHuggingFaceResolvers([LiteRtLmEngine()])` yields a `LitertlmManifestResolver`, and it is the canonical `const` instance — the package's half of the dedup contract. No `initialize()`, no mock.
- One place I did not land where you asked: the `shared_preferences` dev-dependency stays. The end-to-end test in `live_hugging_face_test.dart` still drives `initialize()` — its token travels through `ServiceRegistry`, whose `initialize` opens SharedPreferences — so the pubspec comment now names that one opt-in suite as the reason. If you would rather that leg call the resolver directly and the dependency go, I can do that here.

Locally on 3.47.1: core 701 passed, this package 106 passed / 5 skipped. The live leg with a token passes sha256/size, every URL, and the end-to-end; its two catalog checks report that the catalog moved since the 09-02 snapshot (3 manifests reshipped, 2 new sarashina2.2 repos), which is the signal it exists for. I'll send that refresh as its own PR after this one lands, unless you'd rather have it here.

Thanks again.

DenisovAV (approved):
The move is right where I wanted it, and I checked it by reading the files rather than the description: the core test drives the real `initialize()` against `_FakeEngine implements InferenceEngineProvider, HuggingFaceResolverSource` with no mention of LiteRtLm, so it pins the contract instead of one engine; the litertlm test is `engineHuggingFaceResolvers([LiteRtLmEngine()])` with no init path and no storage mock.

**On `shared_preferences`: keep it.** My objection was never the dependency — it is dev-only and non-transitive, and core already depends on it directly, so it is in every consumer's graph regardless. The objection was that it was a *symptom* of a test living in the wrong package, and that is now cured. What remains is legitimate: the live end-to-end leg is the only place that exercises registry plus engine-carried resolver plus real network together, and trading that assertion for dependency purity would be a bad trade. Your pubspec comment names the single opt-in suite that needs it and points at core for the rest, which makes the dependency refutable rather than permanent — that is exactly the right shape.

I also ran an independent review pass over the whole diff. It came back clean on the things that matter most for a test-only PR: the sweep's loops and mismatch collection are non-vacuous, the live suite fails on HTTP exceptions and statuses (with a 10-minute test timeout inside the 30-minute workflow timeout), the fakes exercise the real init path, and the workflow's secret handling and failure visibility are sound.

## One thing for the follow-up PR, not this one

`fixtures/README.md` describes `reference_goldens.json` as coming from the hf-to-litertlm reference reader — which is the property the whole sweep rests on, since goldens generated by the resolver under test would make it self-confirming. But the recipe pins **neither a commit of that repository nor a version of the package**, and nothing next to the goldens records what produced them. So the independence is currently an assertion in prose that cannot be audited or reproduced after the fact.

Please pin the reader's revision in the recipe and record it alongside the goldens — a `_generator` key in the JSON, or a line in the README naming the commit the rows came from. Cheap now, unrecoverable later: at the next regeneration nobody, including you in six months, can demonstrate which reader produced a given row.

Two smaller notes in the same area, both for later: the recipe needs a throwaway package and a manual `dart run`, so regeneration cannot be reproduced in CI and depends on whoever has an hf-to-litertlm checkout; and since both the reader and the resolver are yours, matching goldens prove the implementations agree, not that either matches the spec — the live leg's `sha256`/`size_bytes` check against Hugging Face's LFS metadata is the only leg grounded in data neither implementation produced, which makes it more load-bearing than its opt-in status suggests.

## Catalog refresh

Separate PR after this lands, as you proposed — agreed. This one's value is the mechanism; a data refresh is a different kind of change and wants its own review, especially now that a refresh also means regenerating the goldens.

Merging. Thank you for the two rounds and for taking the split seriously.

PR #504, opened 2026-09-10
feat: ship agent skills for consumers of the package

DenisovAV:
The Dart team's [`skills` CLI](https://dart.dev/blog/skills-cli-1-0-bundle-and-distribute-ai-agent-skills-for-your-packages) (publisher `labs.dart.dev`, 1.0.1) lets a package carry instruction files that a user's coding assistant installs with one command:

```bash
dart run skills@ get
```

It scans the dependency tree, offers what each package bundles, and installs into the location each agent expects — Claude Code, Codex, Cursor, Antigravity, Cline, Copilot and OpenCode.

## The three skills

Chosen from what actually generates issues, not from what is easiest to write.

| Skill | Package | Covers |
| --- | --- | --- |
| `flutter-gemma-setup` | core | Core registers **no** engine — `initialize(inferenceEngines: […])` is required. `installModel` defaults `fileType` to `task`, so a `.litertlm` file is routed to MediaPipe unless declared. Platform floors (`minSdk 30` for `.litertlm`, 26 for builtin_ai, iOS 15 vs 16). |
| `flutter-gemma-inference` | core | `maxTokens` is the **context window**, not the reply length — below 1024 it *crashes* `.litertlm` rather than truncating (#318); `maxOutputTokens` is the one that caps output. `Message.isUser` defaults to `false`, and a missing `isUser: true` returns an empty string with no error. Session/model lifetime, tools, thinking, multimodal. |
| `flutter-gemma-speech-stt` | speech | Model choice (moonshine / Whisper / Parakeet), the 16 kHz mono 16-bit PCM input contract, and the Whisper output language — a property of a *transcription*, not of the loaded model. |

Our own `.claude/skills/` (`release`, `build-native`, `review-pr`, `ci-status`, `upgrade-genkit`) stay where they are. Those are maintainer tools; shipping instructions for rebuilding our dylibs to every consumer would be noise.

## The distribution half would have failed silently

`packages/flutter_gemma/.pubignore` carries a blanket `**/*.md` with only README and CHANGELOG exempted. `dart run skills@ get` resolves the package on disk from `package_config.json` and reads `skills/` from **there** — so every `SKILL.md` was being stripped from the archive and the skills would have reached nobody, with no error at any point.

Measured rather than reasoned about:

```
# before
dart pub publish --dry-run | grep -c SKILL.md
  core:   0        ← stripped
  speech: 1        ← speech has no such rule

# after
  core:   2
  speech: 1
```

## Verified end to end

Not against the spec — against the CLI, from the example app:

```
$ dart run skills@ get --all --agent claude
  [claude] Installed flutter-gemma-setup
  [claude] Installed flutter-gemma-inference
  [claude] Installed flutter-gemma-speech-stt
Installed 3 skill(s) for claude at .claude/skills.
```

Then `skills@ remove --all` to clean up, so nothing from the test run is in this diff.

## Versions

`flutter_gemma` 1.8.0 → **1.8.1**, `flutter_gemma_speech` 0.5.0 → **0.5.1**. `SKILL.md` ships inside the archive, so without a release the files exist only in the repo.

Both READMEs now mention the command — a mechanism nobody knows about is not a mechanism.


DenisovAV:
Opened against the wrong branch — the name `feat/agent-skills` already existed on origin from earlier `flutter_gemma_agent` work (a different meaning of "skills"), so my push was rejected as non-fast-forward and this PR picked up that branch's 100 files instead. Reopening from a branch of my own; the pre-existing branch is untouched.

PR #506, opened 2026-09-10
Add flush(), so a qdrant index survives the process (#492)

### LINKED ISSUE #492: flutter_gemma_rag_qdrant: EdgeShard.flush() is never called, so an index does not survive the process
**Package:** `flutter_gemma_rag_qdrant` 1.3.0 with `qdrant_edge` 0.8.0-dev.3
**Platforms tested:** Android arm64 (two devices, both Android 16, f2fs, 4 KiB pages), Windows

Two separate findings. The first has a fix; the second I could not explain.

### 1. `EdgeShard.flush()` is never called, so nothing survives the process

`grep -rn flush` over the published package returns nothing. Points added via `addDocument` / `addDocumentWithEmbedding` stay in the in-RAM segment, so an index built in one session is gone in the next and the corpus is re-embedded on every launch.

Measured by writing 143 documents (dim 768), then reopening in a **new process**:

| shutdown before exit | documents after reopen |
|---|---|
| nothing | **0** |
| `flush()` | 143 |
| `close()` | 143 |
| `flush()` + `close()` | 143 |
| `flush()` then `exit(0)` (no Rust drop) | 143 |

Adding the method locally is enough:

```dart
// QdrantEdgeClient
Future<void> flush() async {
  _checkOpen();
  try {
    _shard.flush();
  } catch (e) {
    _rethrow(e);
  }
}

// QdrantVectorStore
Future<void> flush() => _serializeLifecycle(_flush);

Future<void> _flush() async {
  final c = _client;
  if (c == null) return;
  await c.flush();
}
```

Exposing `flush()` on `QdrantVectorStore` would be enough for callers to persist once after a bulk index. Note `close()` calls `_shard.unload()` (`qdrant_edge_client.dart:307`), which the SDK declares separately from `flush()`.

### 2. With flush, one device still cannot reopen a valid shard

Same APK, same code, 143 synthetic documents, `flush()`, then a new process:

* SM-M336BU (Android 16, arm64-v8a, f2fs, PAGE_SIZE 4096) — reopens, 143 documents.
* SM-S948U1 (Android 16, arm64-v8a, f2fs, PAGE_SIZE 4096) — fails every time:

```
QdrantException: OperationExceptionEdgeException(Service runtime error:
failed to load segment .../segments/<uuid>: Service runtime error:
Failed to load ID tracker mappings: Service runtime error:
IO Error: No such file or directory (os error 2))
```

The file it names is present and non-empty:

```
$ adb shell run-as <pkg> ls -l files/<store>/qdrant_edge_v1/segments/<uuid>
-rw------- 3360 mutable_id_tracker.mappings
-rw------- 1280 mutable_id_tracker.versions
-rw-------  239 segment.json
-rw-------    5 version.info
drwx------      payload_index  payload_storage  vector_storage
```

**The shard is not corrupt.** Pulled off that device with `adb exec-out run-as <pkg> tar c` and opened on Windows with the same package version, it loads and reports all 143 documents.

Ruled out: corpus size, payload `metadata`, a preceding failed-open/delete/rebuild cycle, a 15 s delay before reopening, accumulated app state (reproduced after a full uninstall), and app-level code (the failing run drives `QdrantVectorStore` directly with synthetic vectors — no models, no plugin).

DenisovAV:
Fixes #492 (first finding).

`flutter_gemma_rag_qdrant` never called `EdgeShard.flush()`. Points added through `addDocument` stayed in the shard's in-RAM segment, so an index built in one session was gone in the next and the corpus was embedded again from scratch on every launch.

@think2execute measured it on two Android devices and named the missing piece precisely — the next open fails with `Failed to load ID tracker mappings`. That turned out to be exactly the right place to look.

## What was actually missing

Writing 20 documents and looking at the segment directory before and after a flush:

| file | before `flush()` | after |
|---|---|---|
| `mutable_id_tracker.mappings` | absent | 420 bytes |
| `mutable_id_tracker.versions` | absent | 160 bytes |

Everything else — vector storage, payload storage, `segment.json` — is on disk well before this. The id tracker is the part that is not, and it is the part the failing open asks for.

## Why `close()` was not already enough

It persists, because it unloads the shard and unload writes too. But that only covers a store the app closes cleanly. An Android app the system kills in the background never gets there, and until now there was no way to persist *without* giving up the store.

## Scope

`flush()` joins the `VectorStoreRepository` contract, not just the qdrant class, with a no-op default body. The SQLite stores are already durable — `sqlite3` autocommits, and the web VFS writes through to OPFS or IndexedDB — so their implementations say so at their own site rather than leaving a caller to guess which backends need the call.

All seven implementations use `implements` rather than `extends`, so the default body is documentation rather than inheritance; each one declares the method.

The qdrant flush runs on the lifecycle lane beside `close` and `clear` — a flush overlapping a close would reach a shard the other call had already unloaded — and is quiet on a store that was never initialized, since callers flush from lifecycle callbacks they cannot make conditional.

## Verified

* 4 new tests in `qdrant_lifecycle_test.dart`, against a real shard rather than a mock
* mutation-checked: dropping the `flush()` call fails the id-tracker test, and only that one
* `flutter_gemma_rag_qdrant` 113/113, `flutter_gemma_rag_sqlite` 121/121, `flutter_gemma` 701/701
* `flutter analyze packages/` — no errors or warnings

The dependency floor moves to `flutter_gemma: ^1.7.3` in both RAG packages, since their `@override flush()` needs the method to exist in the contract.

## Not addressed

The second finding in #492 — one Samsung device failing to reopen a shard that is intact and loads on Windows — is untouched here. It looks like it belongs to `qdrant_edge` or its Rust layer rather than to this package, and it needs its own reproduction.


DenisovAV:
Superseded by #507.

This branch was cut from `feat/codelab-function-calling` instead of `main`, so it carried that PR's 566 files into a 16-file change. #507 is the same fix on a clean base, plus the review findings that came back on this one — most importantly that the web `flush()` was declared a no-op when the IndexedDB VFS does not sync on commit.

Issue #214, opened 2026-04-06
Garbage repeating output on GPU


CorruptionHades:
I compiled the example from this repo to try out the new Gemma 4 e2b on my Samsung Galaxy A55.
However when using GPU inference the output is repeated garbage.
Switching to CPU gives normal output, is way slower however.

Could be related to this
https://github.com/google-ai-edge/LiteRT-LM/issues/1850

<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/6f5dbe40-5154-4111-a740-5984745ac012" />

CorruptionHades:
Seems like this is causing it:

OpenCL sampler not available, falling back to statically linked C API: UNAVAILABLE: ERROR: [third_party/odml/litert_lm/runtime/components/sampler_factory.cc:294]
W/native  ( 4674): └ Could not load shared library libLiteRtTopKOpenClSampler.so: dlopen failed: library "libLiteRtTopKOpenClSampler.so" not found.

DenisovAV:
Should be fixed in 0.15.1 (PR #270 — Android OpenCL sampler dlopen). Please retry on `^0.15.1` and let me know if you can still reproduce — if yes, I'll need more details (full logcat, model URL).

DenisovAV:
Calling this fixed in 0.15.1 (PR #270 — `patchelf --add-needed libLiteRtLm.so` on the Android OpenCL sampler), confirmed by my own retest on Pixel 8 and other reporters. 0.15.2 just landed too with the unified embedding rewrite.

If you can still reproduce garbage GPU output on `^0.15.1` (ideally 0.15.2), please reopen with the model URL and full logcat — I'll dig in. Sorry it took a couple of releases to nail down 🙏

CorruptionHades:
I tested the new version with the example and now just get different garbage so doesnt seem to be fixed. Ill look at the logs later

DenisovAV:
Thanks for retesting. "Different garbage" instead of the old `OpenCL sampler not found` means the 0.15.1 sampler fix worked but there's a second issue underneath — likely the same upstream GPU-decode bug as [LiteRT-LM#1850](https://github.com/google-ai-edge/LiteRT-LM/issues/1850) (`clEnqueueNDRangeKernel - Invalid command queue` on Gemma 4 GPU decode, confirmed on a few Adreno devices).

[...]

PR #520, opened 2026-09-17
F32 Fixes scrambled digits on GPU

callmephil:
## The Bug

Gemma 4 (`.litertlm`) can incorrectly copy digits from prompts longer than ~2,000 tokens when running on the native GPU backend.

For example, given `2026/06/23`:

* E4B → `20226/12/17`
* E2B → `2026/12/7`

The issue reproduces consistently on:

* M3 Max (Metal)
* Snapdragon 8 Elite (Adreno/OpenCL)
* iPhone 11 (Metal)

Shorter prompts and the Web runtime return the correct values.

### Root Cause

LiteRT-LM uses FLOAT16 activations for GPU inference by default, and the published Gemma 4 models also request FP16 through `prefer_activation_type`.

Using FLOAT32 activations fixes the digit scrambling.

`flutter_gemma` previously had no way to request a different activation type.

---

## The Fix

Add an optional `activationDataType` to `getActiveModel` and `createModel`:

```dart
final model = await FlutterGemma.getActiveModel(
  preferredBackend: PreferredBackend.gpu,
  activationDataType: ActivationDataType.float32,
);
```

Supported values mirror LiteRT-LM:

```dart
enum ActivationDataType {
  float32,
  float16,
  int16,
  int8,
}
```

The value is passed through `RuntimeConfig` to the LiteRT-LM C API before engine creation.

* `null` preserves existing behavior.
* Changing the value causes the cached model to reload.
* Native LiteRT-LM engines use the setting; Web and MediaPipe ignore it.

### Performance

FLOAT32 significantly increases prompt prefill time but has little effect on decode speed.

For a ~3,100-token prompt:

* Snapdragon 8 Elite E4B: ~3.2 s → ~9.8 s
* iPhone 11 E2B: ~9 s → ~23–37 s
* M3 Max: ~1.2–1.4× slower

The default remains unchanged.

---

## Tests

Added unit tests for:

* Activation data type values and engine arguments
* RuntimeConfig propagation
* Cached-model reload behavior

Added an integration test that reproduces the Gemma 4 digit-scrambling bug using a deterministic ~3,100-token prompt.

Results:

* **Android / S25 Ultra / E4B:** 3/3 incorrect by default, 3/3 correct with FLOAT32.
* **iOS / iPhone 11 / E2B:** 3/3 scrambled by default; none with FLOAT32.
* **macOS / M3 Max / E2B + E4B:** scrambling reproduced by default and fixed with FLOAT32.
* **Web / Chrome 153:** no scrambling with the default configuration.

Also verified:

```text
flutter analyze packages/ --no-fatal-infos
tool/test_all.sh
dart tool/check_skills.dart
skills_lint
```

all pass on Flutter 3.47.2.

---

## Documentation

Updated:

* README
* `troubleshooting.md`
* `litertlm.md`
* Inference skill

---

## Related Issues

Related bugs
[google-ai-edge/LiteRT-LM#3012](https://github.com/google-ai-edge/LiteRT-LM/issues/3012): Gemma 4 E4B on an Adreno 750 writes 202026 for 2026 in ~2k-token prompts. This is the same bug.
[google-ai-edge/LiteRT-LM#2814](https://github.com/google-ai-edge/LiteRT-LM/issues/2814): the same digit scrambling on Metal (iPhone and Mac), fixed by the same FLOAT32 call.
[google-ai-edge/LiteRT-LM#2992](https://github.com/google-ai-edge/LiteRT-LM/issues/2992): FP16 gives garbage on NVIDIA. A comment there fixes an Adreno case with prefer_activation_type = "fp32".
[google-ai-edge/LiteRT-LM#2249](https://github.com/google-ai-edge/LiteRT-LM/issues/2249): Gemma 4 E2B on a Mali GPU doubles digits ("world cup 2022" → world cup 202222). The issue asks for a repetition penalty.
[google-ai-edge/LiteRT-LM#2637](https://github.com/google-ai-edge/LiteRT-LM/issues/2637) (closed): on a Windows GPU, numbers in prompts over 2,000 tokens repeat digits (387 → 3887).
[google-ai-edge/LiteRT-LM#2629](https://github.com/google-ai-edge/LiteRT-LM/issues/2629): open PR for a LITERT_LM_FORCE_F32 environment variable in the C API. The command-line version (#2626) was reverted.
[google-ai-edge/LiteRT-LM#3577](https://github.com/google-ai-edge/LiteRT-LM/issues/3577): GPU output differs from CPU on Adreno for a MiniCPM bundle that already runs at FLOAT32. That is a different bug.
#214 (closed): garbage on a Galaxy A55's Xclipse GPU, from Samsung's OpenCL driver. That is a different bug.

---

## Type of Change

* [ ] Bug fix
* [x] New feature
* [x] Documentation update
* [ ] Refactoring
* [x] Test addition

</code></pre><p>🧑‍💻 Generated with Claude Code</p></body></html>


DenisovAV:
Thanks for the thorough PR — the fix is real, and it lines up with upstream: LiteRT-LM added the same `activationDataType` option to its Kotlin and Swift SDKs this week. I'll take care of the version bumps. A few changes before merge:

**1. Please drop `int16` and `int8`** and keep `ActivationDataType { float32, float16 }`. At our LiteRT-LM pin they don't behave like float16: the GPU delegate compiles at fp16, but `use_fp16_precision` is only true for FLOAT16 (`llm_litert_compiled_model_executor.cc:1774`), so the attention mask gets the fp32 fill value (`-0.7 * FLT_MAX` instead of `-45824`) and the logits buffer stays fp32. Upstream's Kotlin config also says unsupported types can fail engine init, which here becomes a silent fallback to CPU. They're easy to add back once someone verifies them on a device.

**2. Docs**
- Add "not on web" to the README block and to `troubleshooting.md` — the web engine ignores the value, and those two pages are where someone with wrong digits will look.
- List ONNX and built-in AI among the engines that ignore it, and say it applies to the text decoder only (the vision and audio encoders keep the model's own setting).
- Mention that float32 needs more GPU memory, and that if the GPU engine can't be created the model falls back to CPU without an error — so apps should check `model.activeBackend == PreferredBackend.gpu` after loading.

**3. Integration test**
- **An answer with no figures passes.** `_askAll` only records answers that contain a figure missing from the log, so a refusal or "not found" counts as clean, and `missing` never fails the float32 test. Please fail when an answer contains none of the expected figures.
- **Nothing shows that float32 actually took effect.** On a GPU without the bug, or if the setter call were removed, the float32 test still passes. The FFI client already enables benchmarking (`litert_lm_client.dart:728`), so `session.getSessionMetrics()` reports time to first token on native. Printing it next to each answer for both runs would show the prefill slowdown that proves fp32 ran. A hard ratio assert would be flaky (the difference on an M3 Max is small), so printing is enough — or assert it behind a `--dart-define` on the devices where you reproduced the bug.
- **Android model:** your Android result is with E4B, but the test defaults to E2B. Please note in the header that Android needs `--dart-define=GEMMA4=E4B` for a meaningful run.
- **Web:** please use `flutter drive --driver=test_driver/integration_test.dart --target=integration_test/activation_data_type_test.dart -d web-server`, as in `litertlm_web_test.dart` — `-d chrome` hangs there on Flutter 3.47.2, and `flutter run` reports no pass/fail. Note too that `integrationDriver` gives up after 20 minutes by default while the tests allow 30, so a first web run that downloads the 2 GB model can hit that limit.

callmephil:
> Thanks for the thorough PR — the fix is real, and it lines up with upstream: LiteRT-LM added the same `activationDataType` option to its Kotlin and Swift SDKs this week. I'll take care of the version bumps. A few changes before merge:
> 
> **1. Please drop `int16` and `int8`** and keep `ActivationDataType { float32, float16 }`. At our LiteRT-LM pin they don't behave like float16: the GPU delegate compiles at fp16, but `use_fp16_precision` is only true for FLOAT16 (`llm_litert_compiled_model_executor.cc:1774`), so the attention mask gets the fp32 fill value (`-0.7 * FLT_MAX` instead of `-45824`) and the logits buffer stays fp32. Upstream's Kotlin config also says unsupported types can fail engine init, which here becomes a silent fallback to CPU. They're easy to add back once someone verifies them on a device.
> 
> **2. Docs**
> 
>     * Add "not on web" to the README block and to `troubleshooting.md` — the web engine ignores the value, and those two pages are where someone with wrong digits will look.
> 
>     * List ONNX and built-in AI among the engines that ignore it, and say it applies to the text decoder only (the vision and audio encoders keep the model's own setting).
> 
>     * Mention that float32 needs more GPU memory, and that if the GPU engine can't be created the model falls back to CPU without an error — so apps should check `model.activeBackend == PreferredBackend.gpu` after loading.
> 
> 
> **3. Integration test**
> 
>     * **An answer with no figures passes.** `_askAll` only records answers that contain a figure missing from the log, so a refusal or "not found" counts as clean, and `missing` never fails the float32 test. Please fail when an answer contains none of the expected figures.
> 
>     * **Nothing shows that float32 actually took effect.** On a GPU without the bug, or if the setter call were removed, the float32 test still passes. The FFI client already enables benchmarking (`litert_lm_client.dart:728`), so `session.getSessionMetrics()` reports time to first token on native. Printing it next to each answer for both runs would show the prefill slowdown that proves fp32 ran. A hard ratio assert would be flaky (the difference on an M3 Max is small), so printing is enough — or assert it behind a `--dart-define` on the devices where you reproduced the bug.
> 
>     * **Android model:** your Android result is with E4B, but the test defaults to E2B. Please note in the header that Android needs `--dart-define=GEMMA4=E4B` for a meaningful run.
> 
>     * **Web:** please use `flutter drive --driver=test_driver/integration_test.dart --target=integration_test/activation_data_type_test.dart -d web-server`, as in `litertlm_web_test.dart` — `-d chrome` hangs there on Flutter 3.47.2, and `flutter run` reports no pass/fail. Note too that `integrationDriver` gives up after 20 minutes by default while the tests allow 30, so a first web run that downloads the 2 GB model can hit that limit.

Thanks, I'll update the PR asap.

PR #521, opened 2026-09-17
refactor(builtin_ai): rebuild on flutter_local_ai, drop the native layer

kekko7072:
This is the dependency direction you asked for: **`flutter_local_ai` stays standalone and never depends on flutter_gemma**, and `flutter_gemma_builtin_ai` becomes a thin adapter on top of it. It replaces builtin_ai's native layer with a dependency on [`flutter_local_ai`](https://pub.dev/packages/flutter_local_ai) (now published, `0.1.0`), keeping only the part that is flutter_gemma's business: the mapping onto `InferenceEngineProvider` / `InferenceModel` / `InferenceModelSession`.

Your three reasons all hold, and this PR is shaped around them:

1. **No flutter_gemma in the graph of a Gemma-free app.** `flutter_local_ai` depends on `flutter`, `plugin_platform_interface` and `flutter_web_plugins` — nothing else. No native plugin on all platforms, no `background_downloader`, no `large_file_handler`, and its floor is Dart 3.8 / Flutter 3.32, not 3.12 / 3.44.
2. **The adapter lives in your repo, so interface changes stay in one PR.** It imports `flutter_gemma/core/...` freely (engine, model, session, resolver) — exactly the internal paths that would break across a repo boundary. A file move on your side is now a same-PR fix.
3. **No migration for current users.** Same package, same imports, same names.

## What changes

`flutter_gemma_builtin_ai` is **pure Dart** from 0.3.0 — no `flutter: plugin:` block, no Kotlin, no Swift, no pigeon. The engine, model, session and model specs (Windows and web included) map `flutter_local_ai` onto flutter_gemma's contracts.

**Public API is unchanged.** `BuiltInAi`, `BuiltInAiEngine`, `BuiltInAiModels`, `BuiltInAiHuggingFaceResolver`, `BuiltInAiAvailability` and `BuiltInAiUnavailableException` keep their names, members and signatures; the engine and resolver keep the identity strings `'BuiltInAI'` and `'builtin-ai-huggingface'`. The two availability types are now typedef aliases of their `LocalAi` counterparts, which declare the same seven values in the same order — verified against `origin/main`, so existing exhaustive `switch` statements keep compiling. `toString()` prints the new name.

Two names are added, not changed: `BuiltInAiModel` and `BuiltInAiSession`, exported only for their `localAiModel` / `localAiSession` getters.

## What consumers gain

- **Windows**, through Windows AI Foundry / Phi Silica.
- **One web arm** instead of two implementations of the same Prompt API.
- **Native tool calling and schema-constrained output** — no slot on `InferenceModelSession`, reachable through `BuiltInAiModel.localAiModel` / `BuiltInAiSession.localAiSession`. As you said, the native Apple tools stay available through `localAiModel` and flutter_local_ai's own API.

flutter_gemma's tools are still deliberately **not** forwarded to flutter_local_ai's native tool API: `InferenceChat` owns function calling and weaves declarations into the prompt, so handing the same tools to the native runner would run two tool loops for one turn.

## What to check before release

- **No longer a plugin** — registrants and `Podfile.lock` regenerate on the next `pub get`. CI running a frozen `pod install --deployment` needs re-locking.
- **macOS floor rises 10.15 → 12.0** (flutter_local_ai's podspec and `Package.swift`). A macOS 11 app fails at `pod install` with a CocoaPods resolution error naming the pod, so the CHANGELOG calls it out. iOS is unaffected in practice: flutter_local_ai builds from 13.0, and core flutter_gemma still requires 15.0.
- **Vision is refused at model creation**, not when the first image is added. An app passing `supportImage: true` unconditionally on a backend with no multimodal path now fails at startup instead of mid-turn. This is the one runtime behaviour change behind the "no code migration" headline.

## Verification

- `flutter analyze` on `flutter_gemma_builtin_ai`: **no issues**.
- `flutter test`: **54 tests pass.** The VM suites that mocked the deleted pigeon channel by name are rewritten against flutter_local_ai's published test double, including the two hang regressions from the Firebase Test Lab fix. The browser suites are dropped here because that behaviour now lives in flutter_local_ai, which tests it itself.
- `flutter analyze` on `packages/flutter_gemma/example` (which depends on this package): **0 errors, 0 warnings** — the remaining `info` lints are pre-existing.
- Resolution verified against the **published** `flutter_local_ai 0.1.0` from pub.dev, not a path dependency. The lockfile diff is the eight-line `flutter_local_ai` entry and nothing else; I reverted the unrelated version churn my local SDK introduced.
- Rebased onto current `main` (`4044852`). The docs conflicts are merged rather than overwritten — your Edge / Phi-4-mini findings and the `dart run skills@ get --all` section are preserved, with the Windows row added alongside. The Edge notes stay accurate: flutter_local_ai's web arm gates on `'LanguageModel' in self`, not on a browser name, and keeps the same `measureContextUsage` → `measureInputUsage` fallback.

Not verified by me: **on-device runs**. Happy for you to take it through Android, iOS, macOS and web on your devices before release, as you suggested — and the Windows path in particular still needs device validation.

Glad to go through it together next week.


kekko7072:
HI @DenisovAV i opened the PR with the refactor for adding the flutter_local_ai as base, also there are some new features related to the local OS API.

